PR #9, opened 2017-11-28
Fix 0 change output

lvaccaro:
The transaction has an 0BTC change amount, in the following condition:

current wallet balance == amount to spend && feeAmount == minAmount 

In this case, the following condition is TRUE (at line 645) and it adds the 0BTC change output:
```
else if (transaction && balance - (amount + feeAmount) >= minAmount) { // add change output
```
Note: same problem if the balance satisfied the following condition:
balance == minAmount + amount + feeAmount


voisine:
I have no issue with this patch, it has the same effect as adding a satoshi to minAmount, however it will only result in a zero output in the cases you mentioned if minAmount is 0

PR #16, opened 2018-02-23
Spell checker (via ispell)

siginfo:
I ran ispell on macOS and fixed some typos. I was able to run it through the compilation phase by adding cmake infrastructure - but hopefully nothing will break as the changeset is only touching comments and not the actual code.

### DIFF
--- a/BRSet.c
+++ b/BRSet.c
@@ -68,7 +68,7 @@ static void _BRSetInit(BRSet *set, size_t (*hash)(const void *), int (*eq)(const
     set->eq = eq;
 }
 
-// retruns a newly allocated empty set that must be freed by calling BRSetFree()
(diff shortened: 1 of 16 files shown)


BRSet.c
```diff
@@ -207,7 +207,7 @@ void *BRSetGet(const BRSet *set, const void *item)
     return t;
 }
 
-// interates over set and returns the next item after previous, or NULL if no more items are available
+// integrates over set and returns the next item after previous, or NULL if no more items are available
```
losh11: iterates*

siginfo: Good catch - fixed

EBGToo:
We never got to this (obviously); and have moved on.  Closing.

PR #36, opened 2019-01-29
Code Cleanup

arubinofaux:
Given that "there’s nothing actually at http://seed.breadwallet.com  it was used a few years back to evaluate whether or not hosting our own nodes would help with SPV syncing performance."  https://twitter.com/zadikafka/status/1089352887795400705 and to prevent this random DNS server to ever include any Bitcoin node potentially opening users wallets to be used as test data without their concern.

I've removed `testnet-seed.breadwallet.com.` and `seed.breadwallet.com.` from the DNS seed list.

### DIFF
--- a/BRChainParams.h
+++ b/BRChainParams.h
@@ -49,12 +49,12 @@ typedef struct {
 } BRChainParams;
 
 static const char *BRMainNetDNSSeeds[] = {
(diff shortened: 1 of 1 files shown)


EBGToo:
Hi.  Yes, those seed files, for mainnet and testnet, are intentionally empty.  They are placeholders in the event that BRD needs to provide Bitcoin IP addresses.  We won't need to update the code base; only a DNS host or two would need to be updated.

PR #37, opened 2019-02-16
Up Token Gas Limit Default

wjmelements:
#### Changes
Our users are having issues redeeming TrueUSD for USD from Bread Wallet because the default token gas limit is too low.
For example, [this transaction](https://etherscan.io/tx/0xb2aded438eb18524c26c0832016cabaefb6bb62aa37cce65c54577a79d741256) required more than 100 kgas before refund.
This increases the gas limit to a value high enough for TrueUSD redemptions.
Reviewers @EBGToo

### DIFF
--- a/ethereum/BREthereumToken.h
+++ b/ethereum/BREthereumToken.h
@@ -32,7 +32,7 @@
 
 // For tokenBRD define defaults for Gas Limit and Price.  These are arguably never up to date
 // and thus should be changed in programmatically using walletSetDefaultGas{Price,Limit}().
-#define TOKEN_BRD_DEFAULT_GAS_LIMIT  92000
+#define TOKEN_BRD_DEFAULT_GAS_LIMIT  112000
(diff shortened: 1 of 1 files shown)


EBGToo:
Hi.  The default gasLimit is used to configure each wallet.  See `walletCreateDetailed()`.  After that initial configuration, a wallet's gasLimit can be changed to a value specific to the wallet's token.  Thus if TrueUSD needs more than 92,000 gas, then the TrueUSD wallet can be set to use 112000.

See `ethereumWalletSetDefaultGasLimit()` and the underlying `walletSetDefaultGasLimit()`.  

Is this not sufficient for your needs?

Cheers.

wjmelements:
We no-longer require this after https://etherscan.io/tx/0x2f05475a016ad0f910dabc95304968d8c4fe8283dddc2f7bc73cd864b69cdac8
92k is now more than enough for any of our transfers.
I'll look into configuring a custom limit, as per your recommendation.

PR #47, opened 2019-05-22
CORE-325: Expose `ewmTransferGetOriginatingTransactionHash()` via JNI

pbudelli:


EBGToo:
This change is a 'work around'; it exposes the specific function to allow a 'Log Transfer' to be associated, via the hash, with the 'Transaction Transfer' (that originated the log).  A better approach would be a) to associate the two transfers and then the 'Transaction Transfer' hash could be displayed in the UI or b) make all associated transfers (Log, Transaction, ...) share the same identifier.  (Probably there is a better 'c' option...)

Option 'b' has a issue in that transfers won't have globally unique identifiers.  If that is the case, then the scope of uniqueness needs to be clarified, for example, 'a wallet'.  But also note, that in general one transaction can generate multiple log transfers but that for ERC20 smart contracts, there is only one log event.  It might be a bigger issue if multiple logs existed for the same transaction.

EBGToo:
I confirmed that the change builds (and didn't prior to the 'forced push change') and that the exceptional minimal, insufficient unit tests (that don't apply to this change) still pass.

EBGToo:
Will approve and merge - understanding that the Java interface will be changing in the future for 'Core Generic'.

PR #62, opened 2019-06-17
CORE-378: Improve BRCryptoTransfer implementation

michael-brd:


EBGToo:
What about the Swift too:
```
public private(set) lazy var amountDirected: Amount = {
        switch direction {
        case .sent: return amount.negate
        case .received: return amount
        case .recovered: return Amount.create(integer: 0, unit: unit)
        }
    }()
```
becomes
```
public private(set) lazy var amountDirected: Amount = {
          return Amount (core: cryptoTransferGetAmountDirected (core), unit: wallet.unit, take: false)
    } ()
```

crypto/BRCryptoTransfer.c
```diff
@@ -379,9 +405,20 @@ cryptoTransferGetHash (BRCryptoTransfer transfer) {
...
+        case BLOCK_CHAIN_TYPE_BTC: {
+            BRWallet *wid = transfer->u.btc.wid;
+            BRTransaction *tid = transfer->u.btc.tid;
 
+            uint64_t fee = BRWalletFeeForTx (wid, tid);
+            uint64_t feePerKb = DEFAULT_FEE_PER_KB;
+            if (UINT64_MAX != fee) {
+                // round to nearest satoshi per kb
+                uint64_t size = BRTransactionVSize (tid);
+                feePerKb = ((fee * 1000) + (size / 2)) / size;
```
EBGToo: 👍

PR #65, opened 2019-06-18
CORE-291

EBGToo:
Fix implementations for various BTC + ETH events - specifically CRYPTO_WALLET_EVENT_{BALANCE,FEE_BASIS}_UPDATED.  Add a function to BRWalletManager for updating a BRWallet's feePerKB while also generated a BTC event.

And a simple reformat of BREthereumEWMEvent.c.


Swift/BRCrypto/BRCryptoWalletManager.swift
```diff
@@ -47,22 +47,26 @@ public final class WalletManager: Equatable {
     /// transfer (like BRD transfer => ETH fee)
     public lazy var primaryWallet: Wallet = {
         // Find a preexisting wallet (unlikely) or create one.
-        return Wallet (core: cryptoWalletManagerGetWallet(core),
+        let coreWallet = cryptoWalletManagerGetWallet(core)!
+        return Wallet (core: coreWallet,
                        listener: system.listener,
                        manager: self,
-                       unit: unit,
```
michael-brd: Purely a question but why not have the cryptoWalletGetUnit() call in the ctor?

crypto/BRCryptoWalletManager.c
```diff
@@ -651,8 +676,29 @@ cwmEventTokenAsETH (BREthereumClientContext context,
                     BREthereumToken token,
                     BREthereumTokenEvent event) {
 
-    BRCryptoWalletManager cwm = context;
-    (void) cwm;
+    BRCryptoWalletManager cwm = context; (void) cwm;
+
+    switch (event) {
+        case TOKEN_EVENT_CREATED: {
+            BRCryptoNetwork network = cryptoWalletManagerGetNetwork (cwm);
```
michael-brd: Is network being leaked here?

crypto/BRCryptoWalletManager.c
```diff
@@ -651,8 +676,29 @@ cwmEventTokenAsETH (BREthereumClientContext context,
...
+
+            // A token was created.  We want a corresponding EWM wallet to be created as well; it
+            // will be created automatically by simply 'pinging' the wallet for token.  However,
+            // only create the token's wallet if we know about the currency.
+
+            if (NULL != cryptoNetworkGetCurrencyForCode (network, tokenGetSymbol(token)))
```
michael-brd: Is currency being leaked here?

crypto/BRCryptoWalletManager.c
```diff
@@ -555,13 +568,33 @@ cwmWalletEventAsETH (BREthereumClientContext context,
             if (NULL == wallet) {
                 BREthereumToken token = ewmWalletGetToken (ewm, wid);
```
michael-brd: Are we leaking anything here (currency, unit)?

crypto/BRCryptoWalletManager.c
EBGToo: Fixed.

crypto/BRCryptoWalletManager.c
EBGToo: Fixed.

Swift/BRCrypto/BRCryptoWalletManager.swift
EBGToo: Fixed.

crypto/BRCryptoWalletManager.c
EBGToo: I don't think so.  The `currency` and `unit` values are explicitly given.  The `wallet` value is taken enough times such that the last event callback has the wallet taken once.  (event handler is responsible to give all arguments)

EBGToo:
@michael-brd I changed the BRCryptoCurrencyCreate() function to add an 'address/issuer' field.  Can you update the Java on this branch?

EBGToo:
@michael-brd Please give the last two commits the once over.  And the IOS CryptoDemo is back to a) updating the wallet balance and b) dynamically inserting transfers as they arrive while viewing a wallet.

PR #66, opened 2019-06-19
CORE-381: For ethereumTestnet install proper tokens (for BRD and TST).

EBGToo:
For Ethereum, ensure that installed tokens are consistent w/ the network.  Specifically, the 'getTokens' BRD endpoint only provides 'mainnet' tokens.


EBGToo:
I considered putting it in the callback but took the shortcut to avoid both a Swift and Java change.  For ow let's leave it in the C and see how it evolves.  

Slightly related, some current test code installs a TST token on 'mainnet' (based on a compiler flag) that doesn't come from the BRD endpoint.  Might consider adding in TST in that case.

PR #67, opened 2019-06-19
CORE-376: Change CWM callbacks to only include primitives or "crypto" data structures

michael-brd:


Java/CoreNative/src/main/java/com/breadwallet/corenative/CryptoLibrary.java
```diff
@@ -205,6 +188,11 @@
     void cryptoWalletManagerDisconnect(BRCryptoWalletManager cwm);
     void cryptoWalletManagerSync(BRCryptoWalletManager cwm);
     void cryptoWalletManagerSubmit(BRCryptoWalletManager cwm, BRCryptoWallet wid, BRCryptoTransfer tid, String paperKey);
+    void cwmAnnounceBlockNumber(BRCryptoWalletManager manager, BRCryptoCallbackHandle handle, long blockHeight, int success);
```
michael-brd: Ideally the announce functions would be usable in BTC/ETH/GEN cases.

crypto/BRCryptoWalletManager.c
```diff
@@ -33,6 +32,27 @@
 static void
 cryptoWalletManagerRelease (BRCryptoWalletManager cwm);
 
+typedef enum  {
+    CWM_CALLBACK_TYPE_BTC_GET_BLOCK_NUMBER,
+    CWM_CALLBACK_TYPE_BTC_GET_TRANSACTIONS,
+    CWM_CALLBACK_TYPE_BTC_SUBMIT_TRANSACTION,
+
+    CWM_CALLBACK_TYPE_ETH_GET_BALANCE
+} BRCryptoCWMCallbackType;
```
michael-brd: Open to suggestions on the naming. BRCryptoCWMClientCallbackType? Too wordy?

crypto/BRCryptoWalletManager.c
```diff
@@ -33,6 +32,27 @@
...
+
+struct BRCryptoCWMCallbackRecord {
```
michael-brd: Open to suggestions on the naming. BRCryptoCWMClientCallbackRecord?

crypto/BRCryptoWalletManager.c
```diff
@@ -66,18 +86,43 @@ static void
 cwmGetBlockNumberAsBTC (BRWalletManagerClientContext context,
                         BRWalletManager manager,
                         int rid) {
-    BRCryptoWalletManager cwm = context;
-    cwm->client.btc.funcGetBlockNumber (cwm->client.context, manager, rid);
+    BRCryptoWalletManager cwm = cryptoWalletManagerTake (context);
```
michael-brd: General pattern is that before invoking the callback, we create a callback record that tracks any state needed in the associated "announce" function to perform its task.

crypto/BRCryptoWalletManager.c
```diff
@@ -66,18 +86,43 @@ static void
...
 static void
 cwmGetTransactionsAsBTC (BRWalletManagerClientContext context,
-                         BRWalletManager btcManager,
+                         BRWalletManager manager,
                          uint64_t begBlockNumber,
                          uint64_t endBlockNumber,
                          int rid) {
-    BRCryptoWalletManager cwm = context;
-    cwm->client.btc.funcGetTransactions (cwm->client.context, btcManager, begBlockNumber, endBlockNumber, rid);
+    BRCryptoWalletManager cwm = cryptoWalletManagerTake (context);
```
michael-brd: This function is the tricky one since the Java (and Swift) currently "loop", announcing transactions, discovering addresses, repeatedly.

I think that behaviour (the looping nature of the get transaction calls) needs to be in BRWalletManager.

crypto/BRCryptoWalletManager.c
```diff
@@ -1233,6 +1309,86 @@ cryptoWalletManagerSubmit (BRCryptoWalletManager cwm,
     }
 }
 
+extern void
```
michael-brd: General pattern is inline with current BWM and EWM. Small departure in that every callback would be expected to call into the associated announce function, even on failure, so that the record could be free'd up.

Java/CoreCrypto/src/main/java/com/breadwallet/corecrypto/WalletManager.java
```diff
@@ -763,158 +765,133 @@ private void handleTransferDeleted(CoreBRCryptoWallet coreWallet, CoreBRCryptoTr
...
     // ETH client
 
-    private void ethGetBalance(Pointer context, BREthereumEwm ewm, BREthereumWallet wid, String address, int rid) {
-        Log.d(TAG, "BREthereumClientHandlerGetBalance");
+    private void ethGetEtherBalance(Pointer context, BRCryptoWalletManager manager, BRCryptoCallbackHandle handle, String networkName, String address) {
+        Log.d(TAG, "BRCryptoCWMEthGetEtherBalanceCallback");
 
-        BREthereumNetwork network = ewm.getNetwork();
-        String networkName = network.getName().toLowerCase();
+        // TODO(discuss): Do we actually need a rid here? Can it just be random? Does it need to tie into our C rid?
```
michael-brd: Big question here; does the RID matter? Does it need to be the same RID as what the Core has?

Java/CoreCrypto/src/main/java/com/breadwallet/corecrypto/WalletManager.java
```diff
@@ -763,158 +765,133 @@ private void handleTransferDeleted(CoreBRCryptoWallet coreWallet, CoreBRCryptoTr
...
+    private void ethGetTokenBalance(Pointer context, BRCryptoWalletManager manager, BRCryptoCallbackHandle handle, String networkName, String address, String tokenAddress) {
+        Log.d(TAG, "BRCryptoCWMEthGetTokenBalanceCallback");
 
-                @Override
-                public void handleError(QueryError error) {
-                    Log.e(TAG, "BREthereumClientHandlerGetBalance: asTok failed", error);
-                }
-            });
-        }
+        // TODO(discuss): Do we actually need a rid here? Can it just be random? Does it need to tie into our C rid?
```
michael-brd: Same RID question as in the getBalanceAsEth()

crypto/BRCryptoWalletManager.h
```diff
@@ -120,10 +120,29 @@ extern "C" {
 
     // MARK: Client
 
+    // TODO(discuss): What do we want to do for naming here?
```
michael-brd: Open to suggestions on the naming

PR #68, opened 2019-06-20
CORE:385: Add feeBasis functions to Ripple

carlcherry:
Ripple transaction fees need to be calculated
based on a base fee plus other parameters
1. Add support in the transaction to store the fee basis,
   calculate and store the actual fee, return both fee basis
   and fee to caller
2. Add support in the wallet to store/retrieve the default fee basis
3. Update unit tests to support new funtionality


EBGToo:
Thanks.  Reviewed and confirmed unit tests pass.

PR #76, opened 2019-07-03
CORE-393: BTC transaction submission and memory management improvement

michael-brd:


Java/CryptoDemo/src/main/java/com/breadwallet/cryptodemo/TransferListActivity.java
```diff
@@ -158,6 +161,36 @@ public Void visit(TransferDeletedEvent event) {
         });
     }
 
+    private List<? extends Transfer> getTransfers() {
```
michael-brd: Ordering is kind of nasty. Do we want to have one or more comparator functions at the C layer?

bitcoin/BRWalletManager.h
```diff
@@ -98,8 +98,41 @@ bwmAnnounceSubmit (BRWalletManager manager,
...
+     * wallet (via BRWalletRegisterTransaction). Transactions are added when they have
+     * been relayed by the P2P network or synced via BlockchainDB.
+     */
     BITCOIN_TRANSACTION_ADDED,
+
+    /**
+     * For P2P, this event occurs once a transaction has been marked as CONFIRMED/UNCONFIRMED
+     * by the P2P network.
+     *
+     * For API, this event does not occur as transactions are implicitly CONFIRMED when synced.
```
michael-brd: In light of conversations with Sam, this statement and assumption will likely need to be revisited. For now, however, it is an accurate statement from the point of view of how the code behaves.

bitcoin/BRWalletManager.h
```diff
@@ -98,8 +98,41 @@ bwmAnnounceSubmit (BRWalletManager manager,
...
+     */
     BITCOIN_TRANSACTION_UPDATED,
+
+    /**
+     * For P2P, this event occurs once a transaction has been deleted from a wallet
+     * (via BRWalletRemoveTransaction) as a result of an UNCONFIRMED transaction no longer
+     * being visible in the mempools of any of connected P2P peers.
+     *
+     * For API, this event does not occur as transactions are implicitly CONFIRMED when synced.
```
michael-brd: In light of conversations with Sam, this statement and assumption will likely need to be revisited. For now, however, it is an accurate statement from the point of view of how the code behaves.

michael-brd:
Some notes/context on the change:

1. Added callbacks for BITCOIN_TRANSACTION_CREATED and BITCOIN_TRANSACTION_SIGNED.

I used callbacks to mimic the ETH behaviour so as to be able to create corresponding crypto events. In addition, the `cryptoWalletCreateTransfer` occurs at the cryptoWallet level, so by using a callback we can get the event sent to the cryptoWalletManager for propagation to its listener.

2. The BITCOIN_TRANSACTION_CREATED and BITCOIN_TRANSACTION_SIGNED callbacks occur directly in their corresponding functions, rather than using an event.

I don't like this but it is because the transaction memory isn't owned by the BRWalletManager, it is owned by the caller. So we can't safely queue in the event handler, for instance.

We could look at having the BRWalletManager own created/signed but not yet submitted transactions. There is some complexity here; namely that unsigned transactions don't have an identifier (like txHash) so identifying them is difficult.

3. Update `cryptoTransferHasBTC` and `cryptoTransferEqualAsBTC` to not be strictly an identity check.

It now checks for identify only IF the transaction is not signed (i.e. no txHash), otherwise it will use identity or txHash.

This allows our cryptoWallet to hold copies of BTC transactions, rather than the underlying BRWallet's transactions (or "unowned" newly created/signed but not submitted transactions).

bitcoin/BRWalletManager.h
EBGToo: With the addition of these enumeration values, in part to parallel the Ethereum transfer events, are we not well on our way to having generic transfer event types?  And events?

Java/CryptoDemo/src/main/java/com/breadwallet/cryptodemo/TransferListActivity.java
EBGToo: Yes, I think we need comparator functions in the C level - nominally {block number, block transaction index} w/ DESC (it seems) for confirmed transaction but some timestamp (+ nonce) fallback for unconfirmed transactions.  And then some suitable Java/Swift interface.  Could be a wallet default (a little hesitant to go there but if Swift/Java need to store the default themselves, might as well be in the Core) and an override option?

bitcoin/BRWalletManager.c
```diff
@@ -710,18 +710,54 @@ BRWalletManagerScan (BRWalletManager manager) {
...
+                                              manager->wallet,
+                                              transaction,
+                                              (BRTransactionEvent) {
+                                                  BITCOIN_TRANSACTION_SIGNED
+                                              });
+    }
+
+    return r;
+}
+
```
EBGToo: I don't see a problem with the above functions as they are consistent w/ current usage.  But recall we've discussed holding a signed transaction if the transaction can be changed subsequently.  (The change invalidates the signature - seen on ETH transactions).  And how maybe create+sign+submit need to be atomic.

crypto/BRCryptoWalletManagerClient.c
```diff
@@ -441,28 +540,54 @@ cwmTransactionEventAsBTC (BRWalletManagerClientContext context,
...
 
+            cryptoTransferGive (transfer);
             break;
         }
 
         case BITCOIN_TRANSACTION_UPDATED: {
+            // This event occurs when the timestamp and/or blockHeight have been changed.
+            // Regardless of if this was a user created transaction or if it came from a
+            // sync, we are holding a copy of the transaction that BRWalletManager has.
+            // As a result, we need to update our copy of the transaction.
```
EBGToo: Maybe a note in the comment about if there is a blockchain reorg we'll get an UPDATE with a block height of TX_UNCONFIRMED (which I see is handled below).

Java/CryptoDemo/src/main/java/com/breadwallet/cryptodemo/TransferListActivity.java
```diff
@@ -158,6 +161,36 @@ public Void visit(TransferDeletedEvent event) {
...
+                return -1;
+            } else if (oc2.isPresent()) {
+                return 1;
+            } else {
+                return o1.hashCode() - o2.hashCode();
+            }
+        });
+        return walletTransfers;
+    }
+
```
EBGToo: `indexCompare` before `dateCompare`, I think.

Java/CryptoDemo/src/main/java/com/breadwallet/cryptodemo/TransferListActivity.java
michael-brd: CORE-400 created.

Java/CryptoDemo/src/main/java/com/breadwallet/cryptodemo/TransferListActivity.java
michael-brd: Hmm, as a user, I'd be more interested in having things ordered by the timestamp, within a block, than its index in the block. Or am I misunderstanding what that index corresponds to?

EBGToo: If confirmed, the timestamp is the block timestamp and thus multiple transactions in one block will have the same timestamp; it is the index within the block, that distinguishes such transactions.  The timestamp is useful when the block is not confirmed (assuming the timestamp is relevant at that point - like assigned on creation).

bitcoin/BRWalletManager.h
michael-brd: Yes, there is definitely convergence at the ergonomic level. The devil is in the details thought and there are some significant threading/ownership considerations that need to be understood and handled on this front.

Java/CryptoDemo/src/main/java/com/breadwallet/cryptodemo/TransferListActivity.java
michael-brd: Ah, I see. So in that case, given that this logic is within the case where both transfers are included, comparing timestamp is unnecessary entirely. Block and then index is all that is required, correct?

bitcoin/BRWalletManager.c
michael-brd: Great!

crypto/BRCryptoWalletManagerClient.c
michael-brd: Will update.

Java/CryptoDemo/src/main/java/com/breadwallet/cryptodemo/TransferListActivity.java
EBGToo: Yes.

PR #79, opened 2019-07-05
CORE-389: For “bch” use “Bitcoin Cash”

EBGToo:
Reword the builtin blockchain and currencies to use the correct name for BCH


EBGToo:
@michael-brd Needs to be in the Java as well.  Care to change it now on this branch?

EBGToo:
Yes, the JSON is fine.  I can then hard-code it (as I must in case BRD is down/dead) and be 100% confident that the hard-coding is consistent.

michael-brd:
Yeah, changing now.

PR #80, opened 2019-07-08
CORE-402: Separate CWM callbacks from WalletManager instance

michael-brd:


Swift/BRCrypto/BRCryptoWalletManager.swift
```diff
@@ -559,6 +559,9 @@ extension WalletManager {
             funcGetBlockNumber: { (context, cwm, sid) in
                 let manager = Unmanaged<WalletManager>.fromOpaque(context!).takeUnretainedValue()
                 precondition (nil != cwm, "SYS: BTC: GetBlockNumber: Missed {cwm}")
+                defer {
+                    if let cwm = cwm { cryptoWalletManagerGive(cwm) }
+                }
```
EBGToo: FYI: Since the `precondition()` confirms the `cwm` is not `nil`, the `defer` can be `defer { cryptoWalletManagerGive (cwm!) }`

EBGToo: I think CORE-392 will eventually provide `precondition(nil != cwm)` for all/most-of the cases below.

crypto/BRCryptoWalletManagerClient.h
```diff
@@ -45,89 +45,89 @@ extern "C" {
...
     typedef void
         (*BRCryptoCWMEthGetEtherBalanceCallback) (BRCryptoCWMClientContext context,
-                                                  BRCryptoWalletManager manager,
-                                                  BRCryptoCWMClientCallbackState callbackState,
-                                                  const char *network,
-                                                  const char *address);
+                                                  OwnershipGiven BRCryptoWalletManager manager,
+                                                  OwnershipGiven BRCryptoCWMClientCallbackState callbackState,
+                                                  OwnershipKept const char *network,
+                                                  OwnershipKept const char *address);
```
EBGToo: I usually consider  `const char *` as implicitly OwnershipKept but being explicit is better.

Swift/BRCrypto/BRCryptoWalletManager.swift
michael-brd: I added the precondition check throughout, since it should never be null for these callbacks.

After more thought, I've also moved the release of CWM to the completion handler.

crypto/BRCryptoWalletManagerClient.h
michael-brd: Didn't know how far down we wanted to go with this.

Swift/BRCrypto/BRCryptoWalletManager.swift
EBGToo: Good catch on the completion handler.

michael-brd:
@EBGToo, created CORE-407 to track the corresponding Swift change. Once you've approved this PR, I will merge.

PR #82, opened 2019-07-09
CORE-306: Add locks to crypto data structures

michael-brd:


michael-brd:
While adding locks to the crypto data structures, I removed the use of wallets in cryptoTransfer. If you'd prefer, I can split that into a separate change/PR.

crypto/BRCryptoTransfer.c
```diff
@@ -426,11 +411,11 @@ cryptoTransferGetFeeBasis (BRCryptoTransfer transfer) {
...
         }
         case BLOCK_CHAIN_TYPE_ETH: {
-            BREthereumEWM ewm = transfer->u.eth.ewm;
             BREthereumTransfer tid =transfer->u.eth.tid;
 
-            BREthereumGas gas = ewmTransferGetGasLimit (ewm, tid);
-            BREthereumGasPrice gasPrice = ewmTransferGetGasPrice (ewm, tid, WEI);
+            BREthereumFeeBasis feeBasis = transferGetFeeBasis (tid);
+            BREthereumGas gas = feeBasisGetGasLimit (feeBasis);
+            BREthereumGasPrice gasPrice = feeBasisGetGasPrice (feeBasis);
```
EBGToo: Oh, nice!

EBGToo: Probably applies in other places, notably Wallet.

crypto/BRCryptoTransfer.c
```diff
@@ -89,9 +93,43 @@ cryptoTransferCreateAsBTC (BRCryptoCurrency currency,
...
+
+        for (size_t index = 0; index < outputsCount; index++) {
+            if (outputsContain == BRWalletContainsAddress(wid, outputs[index].address)) {
+                BRAddress address;
+                memcpy (address.s, outputs[index].address, sizeof (address.s));
+                transfer->targetAddress = cryptoAddressCreateAsBTC (address);
+            }
+        }
+    }
+
```
EBGToo: There is no `break` in the above?  `transfer->targetAddress` is assigned possibly multiple times?  Ditto `transfer->sourceAddress` above?

crypto/BRCryptoTransfer.c
```diff
@@ -152,28 +180,9 @@ cryptoTransferGetSourceAddress (BRCryptoTransfer transfer) {
...
-            BRTxOutput *outputs      = tid->outputs;
-
-            int outputsContain = (!sent ? 1 : 0);
-
-            for (size_t index = 0; index < outputsCount; index++)
-                if (outputsContain == BRWalletContainsAddress(wid, outputs[index].address)) {
-                    BRAddress address;
-                    memcpy (address.s, outputs[index].address, sizeof (address.s));
-                    return cryptoAddressCreateAsBTC (address);
-                }
```
EBGToo: A break-like `return`.

crypto/BRCryptoWallet.c
```diff
@@ -40,6 +42,8 @@ cryptoWalletRelease (BRCryptoWallet wallet);
 
 
 struct BRCryptoWalletRecord {
+    pthread_mutex_t lock;
+
```
EBGToo: A Wallet-based lock?  I would have thought only a WalletManager-based lock... but a) easy to create locks, b) easy to encapsulate the locks (instead of, for example, needing to pass it down from WalletManager).

crypto/BRCryptoWallet.c
```diff
@@ -191,57 +205,83 @@ cryptoWalletGetBalance (BRCryptoWallet wallet) {
...
-    return NULL;
+    BRCryptoTransfer transfer = NULL;
+    pthread_mutex_lock (&wallet->lock);
+    for (size_t index = 0; index < array_count(wallet->transfers) && NULL == transfer; index++) {
+        if (CRYPTO_TRUE == cryptoTransferHasETH (wallet->transfers[index], eth)) {
+            transfer = cryptoTransferTake (wallet->transfers[index]);
+        }
+    }
+    pthread_mutex_unlock (&wallet->lock);
+    return transfer;
```
EBGToo: Personally, I'd rather `break` from the `for ... if ...` rather than find that `NULL == transfer` condition...  But that is me.

EBGToo:
> While adding locks to the crypto data structures, I removed the use of wallets in cryptoTransfer. If you'd prefer, I can split that into a separate change/PR.

I think I like it in general.  It heads us down the path of caching anything that is static in the BTC+ETH+GEN level?

crypto/BRCryptoTransfer.c
michael-brd: Good catch. Fixed.

crypto/BRCryptoWallet.c
michael-brd: I've added locks to avoid issues with memory corruption within any of the mutable "classes". Did we want to go beyond that? ETH events are single threaded and ideally BTC will be too so it then becomes a question of do we need to lock for logical integrity.

crypto/BRCryptoWallet.c
michael-brd: Compromise; I've switched to breaks for all but the cases where we are returning a boolean (ex: cryptoNetworkHasCurrency).

michael-brd:
> > While adding locks to the crypto data structures, I removed the use of wallets in cryptoTransfer. If you'd prefer, I can split that into a separate change/PR.
> 
> I think I like it in general. It heads us down the path of caching anything that is static in the BTC+ETH+GEN level?

For now, I erred on the side of "lets minimize external dependencies" but we certainly could look to do more caching, assuming performance isn't poor.

crypto/BRCryptoWallet.c
EBGToo: I think it is fine.

PR #83, opened 2019-07-09
CORE-404: Set cryptoNetwork height in C layer

michael-brd:


Swift/BRCrypto/BRCryptoWalletManager.swift
```diff
@@ -109,17 +109,7 @@ public final class WalletManager: Equatable {
...
-            network.height = newValue
-            announceEvent (WalletManagerEvent.blockUpdated(height: newValue))
-            wallets.flatMap { $0.transfers }
-                .forEach {
-                    if let confirmations = $0.confirmationsAt (blockHeight: newValue) {
-                        $0.announceEvent (TransferEvent.confirmation (count: confirmations))
-                    }
-            }
-        }
+        return network.height
```
EBGToo: How is the TransferEvent.confirmation generated now?  Maybe in `walletManagerEventCallback`?

Swift/BRCrypto/BRCryptoWalletManager.swift
```diff
@@ -401,7 +391,9 @@ extension WalletManager {
                                                           event: WalletManagerEvent.syncEnded(error: nil))
 
                 case CRYPTO_WALLET_MANAGER_EVENT_BLOCK_HEIGHT_UPDATED:
-                    manager.height = event.u.blockHeight.value
+                    manager.listener?.handleManagerEvent (system: manager.system,
+                                                          manager: manager,
+                                                          event: WalletManagerEvent.blockUpdated(height: event.u.blockHeight.value))
```
EBGToo: Is this the place to handle the TransferEvent.confirmation event?

Swift/BRCrypto/BRCryptoWalletManager.swift
michael-brd: Confirmation is captured via CRYPTO_TRANSFER_EVENT_CONFIRMED and a corresponding state.

You have a comment to the effect that is CRYPTO_TRANSFER_EVENT_CONFIRMED is unneeded.

Swift/BRCrypto/BRCryptoWalletManager.swift
michael-brd: Confirmation is reflected in the transfer's state. If that's not what you had intended, I would think any event generation would be at the crypto C layer.

Swift/BRCrypto/BRCryptoWalletManager.swift
EBGToo: Nothing is generating `CRYPTO_TRANSFER_EVENT_CONFIRMED` (it seems?) and the listener callback is stubbed out (with the `10` below):
```
                case CRYPTO_TRANSFER_EVENT_CONFIRMED:
                    transfer.listener?.handleTransferEvent (system: manager.system,
                                                            manager: manager,
                                                            wallet: wallet,
                                                            transfer: transfer,
                                                            event: TransferEvent.confirmation(count: 10))
```

Maybe `BRCryptoTransferState` needs a `u.confirmed.height` field and Crypto C generates 'CONFIRMED' as needed (specially when the blockHeight changes)?

Swift/BRCrypto/BRCryptoWalletManager.swift
EBGToo: I don't think confirmations is reflected in the transfer state because confirmations is a combination of transfer state and network block height.  

The original idea was for an App transfer view needing to update the display as blocks arrive and confirmations change.  The view could listen on Transfer events alone (if TransferEvent.confirmations exists) or on Transfer events (for other updates - confirmed, etc) and WalletManager.blockUpdated  events.  I thought listening just on Transfer events when displaying a TransferView was more natural.  Is the Java CryptoDemo displaying 'confirmations'?

michael-brd: The Java has the same behaviour as the Swift algorithm (shown below).

From Swift/BRCrypto/BRCryptoTransfer.swift
```
    /// An optional confirmation.
    public var confirmation: TransferConfirmation? {
        if case .included (let confirmation) = state { return confirmation }
        else { return nil }
    }

    public func confirmationsAt (blockHeight: UInt64) -> UInt64? {
        return confirmation
            .flatMap { blockHeight >= $0.blockNumber ? (1 + blockHeight - $0.blockNumber) : nil }
    }

    /// The confirmations of transfer at the current network `height`.
    public var confirmations: UInt64? {
        return confirmationsAt (blockHeight: wallet.manager.network.height)
    }
```

michael-brd:
* Remove CRYPTO_TRANSFER_EVENT_CONFIRMED from source
* Add comment to Transfer.confirmations code highlighting need to respond to WM block height events to update confirmation count
* Update WalletManager.blockUpdated event docs to highlight that it impacts transfer block height

Swift/BRCryptoDemo/TransferViewController.swift
```diff
@@ -251,9 +251,6 @@ class TransferViewController: UIViewController, TransferListener {
                     break // nearly impossible
                 }
                 break
-            case .confirmation(let count):
-                self.confCountLabel.text = count.description
-                break
             case .deleted:
```
EBGToo: So arguably `TransferViewController` needs to be a `WalletManagerListener` too (and register as such) and handle a BLOCK_HEIGHT_UPDATED event by updating the `confCountLabel`?

Swift/BRCryptoDemo/WalletViewController.swift
```diff
@@ -117,12 +117,6 @@ class WalletViewController: UITableViewController, TransferListener {
                     self.tableView.reloadRows(at: [path], with: .automatic)
                 }
 
-            case .confirmation:
-                if let index = self.transfers.firstIndex (of: transfer) {
-                    let path = IndexPath (row: index, section: 0)
-                    self.tableView.reloadRows(at: [path], with: .automatic)
-                }
-
```
EBGToo: And here too.  The `reloadRows` will change the color indicator from 'yellow' (pending), to 'green' (confirmed/included).

michael-brd: Updated to account for this.

Swift/BRCryptoDemo/TransferViewController.swift
michael-brd: Updated to account for this.

PR #85, opened 2019-07-10
CORE-316: Require BRCryptoAccount to Include a Timestamp

EBGToo:
Consequently requires Swift Account to include a `Date` argument on account creation (and removes the Account setter for the `timestamp` property).  Also updates the test use of CoreTestsConfig to be a '.json' file (instead of a '.plist') file

Needs to be reflected in the Java.  (Which I need to eventually start doing too...)


EBGToo:
> Should cryptoAccountSetTimestamp() be removed from BRCryptoAccount?

Yes, I've got that in my next PR, coming soon.

PR #86, opened 2019-07-11
CORE-401: Account serialization

EBGToo:
Add account serialization.  Once an account is created, based on a `phrase` (aka 12-word BIP-39 'paper key') it can be serialized and then recreated from the serialization.  Doing so avoids using the sensitive 'paper key' whenever the account needs to be created.

We anticipate that that serialization will change; specifically anytime a new blockchain is added.  In such a case the new blockchain's public key needs to be serialized.  For example, when adding XRP (Ripple) the XRP public key must be serialized.  On a change recreating the account from the serialization will fail and the 'paper key' must be used to recreate the account.

### DIFF
--- a/crypto/BRCryptoAccount.c
+++ b/crypto/BRCryptoAccount.c
@@ -31,6 +31,8 @@
 #include "support/BRKey.h"
 #include "ethereum/BREthereum.h"
 
+#define ACCOUNT_SERIALIZE_DEFAULT_VERSION  1
+
 static void
 cryptoAccountRelease (BRCryptoAccount account);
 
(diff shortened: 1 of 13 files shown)


crypto/BRCryptoAccount.c
```diff
@@ -64,28 +98,57 @@ cryptoAccountCreateFromSeedInternal (UInt512 seed,
...
 extern BRCryptoAccount
 cryptoAccountCreate (const char *phrase, uint64_t timestamp) {
     return cryptoAccountCreateFromSeedInternal (cryptoAccountDeriveSeedInternal(phrase), timestamp);
 }
 
 extern BRCryptoAccount
-cryptoAccountCreateFromSeed (UInt512 seed, uint64_t timestamp) {
-    return cryptoAccountCreateFromSeedInternal (seed, timestamp);
-}
+cryptoAccountCreateFromSerialization (const uint8_t *bytes, size_t bytesCount) {
```
michael-brd: We don't currently us bytesCount anywhere. Should we parse this defensively and check that value?

crypto/BRCryptoAccount.h
```diff
@@ -37,25 +37,46 @@ extern "C" {
...
+     *
+     * @param phrase A BIP-32 Paper Key
+     *
+     * @return A UInt512 seed
+     */
     extern UInt512
     cryptoAccountDeriveSeed (const char *phrase);
 
+
+    extern char *
```
michael-brd: What about taking in the length of the word list, to confirm that it is sufficiently sized (BIP39_WORDLIST_COUNT)? And/Or what do you think about documenting this function?

EBGToo:
Modified the serialization to include a checksum (first 2 bytes) and a size (next 4 bytes).  Use these as a sanity check on provided bytes when reconstructing the account.

crypto/BRCryptoAccount.c
EBGToo: See the latest commit.  The serialization now starts with  a checksum and a size.  The bytesCount must be at least large enough for those two and then the bytes will be checked against both the checksum and the size.

PR #88, opened 2019-07-14
CORE-292: Define GEN interfaces; implement for XRP

EBGToo:
Extends BRCrypto*.[ch] with implementations for BLOCK_CHAIN_TYPE_GEN.  Defines generic/BRGeneric*.[ch] and implements those interfaces using the XRP interfaces.

Changes are of five types:

1. Implement BLOCK_CHAIN_TYPE_GEN
2. Add CWM Client Callbacks; implement BRCryptoCWMClientGEN in BRCryptoWalletManager
3. Some minimal code reordering (move BTC function in front of ETH, which is in front of GEN/XRP)
4. Add some needed code to ripple/
5. Clean up some tests.

The bulk of the change is in #1 above.

This is a 'preemptive merge'; although the unit tests pass (now that they are fixed), the XRP changes have not be tested with the BlockchainDB and there are an unknown or two on XRP account serialization (CORE-414)


Swift/BRCrypto/BRCryptoWalletManager.swift
```diff
@@ -931,7 +931,69 @@ extension WalletManager {
...
+
+                print ("SYS: GEN: GetBlockNumber")
+                manager.query.getBlockchain (blockchainId: manager.network.uids) {
+                    (res: Result<BlockChainDB.Model.Blockchain, BlockChainDB.QueryError>) in
+                    res.resolve (
+                        success: { cwmAnnounceGetBlockNumberSuccessAsInteger (cwm, sid, $0.blockHeight) },
+                        failure: { (_) in cwmAnnounceGetBlockNumberFailure (cwm, sid) })
+                }},
+
+            funcGetTransactions: { (context, cwm, sid, address, begBlockNumber, endBlockNumber) in
```
michael-brd: As a follow up to track (i.e. not blocking the PR), we might want address to be a list of addresses.

crypto/BRCryptoAccount.c
```diff
@@ -186,23 +209,35 @@ if (bytesPtr > bytesEnd) return NULL; /* overkill */ \
...
 #undef BYTES_PTR_INCR_AND_CHECK
 }
 
 static void
 cryptoAccountRelease (BRCryptoAccount account) {
-    accountFree(account->eth);  // Core holds???
 //    printf ("Account: Release\n");
+    accountFree(account->eth);
+    gwmAccountRelease(account->xrp);
+
```
michael-brd: Not to block PR, but going forward (maybe tracked by an issue for us to do a pass), what do you think about zeroing (or 0xFF, whatever) out on release? Would help us catch use after free situations.

crypto/BRCryptoWalletManagerClient.c
```diff
@@ -1379,19 +1392,84 @@ cwmGetNonceAsETH (BREthereumClientContext context,
...
+    callbackState->u.genWithTransaction.tid = transfer;
+    callbackState->rid = rid;
+
+    BRGenericHash hash = gwmTransferGetHash (cwm->u.gen, transfer);
+    char *hashAsHex = genericHashAsString (hash);
+//    UInt256 txHash = UInt256Reverse (transaction->txHash);
+//    char *hashAsHex = encodeHexCreate (NULL, txHash.u8, sizeof(txHash.u8));
+
+
+    size_t txLength = 0; // BRTransactionSerialize (transaction, NULL, 0);
```
michael-brd: Even as a placeholder, this will lead to nasty corruption (on Android). Can we put a TODO and put this as 1 at least?

Swift/BRCrypto/BRCryptoWalletManager.swift
EBGToo: CORE-415

crypto/BRCryptoWalletManagerClient.c
EBGToo: CORE-415

crypto/BRCryptoAccount.c
EBGToo: CORE-415

PR #100, opened 2019-07-17
CORE-420: Add Account.validatePhrase

EBGToo:


Swift/BRCrypto/BRCryptoAccount.swift
```diff
@@ -33,6 +33,15 @@ public final class Account {
         return Data (bytes: bytes!, count: bytesCount)
     }
 
+    internal init (core: BRCryptoAccount, uids: String, take: Bool) {
```
michael-brd: Can we move the UIDS down to the C layer (tracked in a separate issue/PR, so as not to gate)? In keeping with our mantra of the C API being a full fledged implementation, I don't think we want anything tracked at the wrapper layer.

EBGToo: Yes, and see CORE-421 (the similar, big trouble-maker).

Swift/BRCrypto/BRCryptoAccount.swift
```diff
@@ -70,27 +79,37 @@ public final class Account {
...
+    ///
+    /// Validate a phrase as a BIP-39 'paper key'; returns true if validated, false otherwise
+    ///
+    /// - Parameters:
+    ///   - phrase: the candidate paper key
+    ///   - words: A local-specific BIP-39-defined array of BIP39_WORDLIST_COUNT words.
+    ///
+    /// - Returns: true is a valid paper key; false otherwise
+    ///
+    public static func validatePhrase (_ phrase: String, words: [String]) -> Bool {
```
michael-brd: Phrase is a string here, but I had thought we were using Data for this (as above)?

Swift/BRCrypto/BRCryptoAccount.swift
```diff
@@ -70,27 +79,37 @@ public final class Account {
...
+        precondition (CRYPTO_TRUE == cryptoAccountValidateWordsList (Int32(words.count)))
```
michael-brd: Should wordsCount be a size_t or at least an unsigned type?

Swift/BRCrypto/BRCryptoAccount.swift
EBGToo: We have another JIRA for this - CORE-399.  Not addressed yet (should be soon; I asked Ehsan about it recently and his said 'yes/okay').

Swift/BRCrypto/BRCryptoAccount.swift
EBGToo: Agreed.

Swift/BRCrypto/BRCryptoAccount.swift
EBGToo: @michael-brd For the Java Data/byte[], do you include the '\0' in the bytes?  [I think BRKeyPubKey, which uses uint8_t*, includes a '\0' in the bytes and in the returned length; in one case I saw `Data` with a '\0' - and then a string conversion produced "...abc\0"]

michael-brd: I append a \0, yes, to ensure that is null terminated.

Swift/BRCrypto/BRCryptoAccount.swift
EBGToo: CORE-432 - UIDS into BRCryptoAccount

PR #102, opened 2019-07-18
CORE-418: Validate an Account Serialization

EBGToo:
Given an Account, check if a pre-existing serialization is from that Account.  Validation is based on having an identical BTC MasterPublicKey.


crypto/BRCryptoAccount.c
```diff
@@ -332,6 +332,42 @@ cryptoAccountSerialize (BRCryptoAccount account, size_t *bytesCount) {
...
+
+    uint8_t *bytesPtr = (uint8_t *) bytes;
+    uint8_t *bytesEnd = bytesPtr + bytesCount;
+
+    size_t chkSize = sizeof (uint16_t); // checksum
+    size_t szSize  = sizeof (uint32_t); // size
+    size_t verSize = sizeof (uint16_t); // version
+    size_t tsSize  = sizeof (uint64_t); // timestamp
+
+    // Skip directly to the BTC MPK
```
michael-brd: Shouldn't we be checking all these fields for validity?

EBGToo: At which point?  Extract the `version`, check if known?  Extract the `timestamp` check if 'recent-y'?

michael-brd: Certainly checksum and version, I would think. What is szSize, just the header size?

michael-brd: Given the context from the issue behind this change, I see now why we don't want to check those fields.

michael-brd: If this is to confirm the same seed material, I think the name could be more reflective of that. I think the assumption someone might have looking at it (from a client POV) is that it will confirm that the serialization is "valid" and thus can be loaded when in reality the version might be unsupported, the checksum could be corrupted, etc.

Approving but food for thought.

EBGToo: CORE-433

PR #103, opened 2019-07-18
CORE-425: Avoid an extra cryptoWalletMangerGive() on GEN option

EBGToo:


crypto/BRCryptoWalletManagerClient.c
```diff
@@ -1437,8 +1437,6 @@ cwmGetBlockNumberAsGEN (BRGenericClientContext context,
     callbackState->rid = rid;
 
     cwm->client.gen.funcGetBlockNumber (cwm->client.context, cwm, callbackState);
```
michael-brd: By not giving up our reference, we're diverging from BTC/ETH. Why is that?

EBGToo: The GEN code should be like the following, mirroring the BTC code (take in client call-out, give on return)?  Is that the concern?
```
static void
cwmGetBlockNumberAsGEN (BRGenericClientContext context,
                        BRGenericWalletManager manager,
                        int rid) {
    BRCryptoWalletManager cwm = cryptoWalletManagerTake (context);

    BRCryptoCWMClientCallbackState callbackState = calloc (1, sizeof(struct BRCryptoCWMClientCallbackStateRecord));
    callbackState->type = CWM_CALLBACK_TYPE_GEN_GET_BLOCK_NUMBER;
    callbackState->rid = rid;

    cwm->client.gen.funcGetBlockNumber (cwm->client.context,
                                        cryptoWalletManagerTake (cwm),
                                        callbackState);

    cryptoWalletManagerGive (cwm);
}
```

EBGToo: The client callout is required to give.

michael-brd: The BTC equivalent (`cwmGetBlockNumberAsBTC`) takes on entry and gives on exit. The GEN one now doesn't. Was trying to understand why the difference.

EBGToo: The difference was this `cryptoWalletManagerTake (cwm)` in the `cwm->client...` invocation.  BTC has that take and the give-on-return, the GEN skipped them both.  I'll parallel the BTC code.

EBGToo: I pushed.  Now like the BTC code.

PR #114, opened 2019-07-25
CORE-429: Provide Crypto Key

EBGToo:


EBGToo:
> Devils advocate; if we are implementing all this in Swift (and Java), should it instead be at the C layer?

CORE-436

PR #115, opened 2019-07-26
CORE-317: Allow `System` to be Reclaimed

EBGToo:
When a system is discarded make sure it is reclaimed (it is).  When a new system is created make sure it can be used (it couldn't - owing to an error in `System.systemExtract()`.  Update the IOS Demo App.

No required Java changes.


Swift/BRCryptoDemo/CoreDemoAppDelegate.swift
```diff
@@ -191,5 +193,28 @@ extension UIApplication {
...
+        app.system.stop()
+        app.summaryController.reset()
+
+        
+        // Assign and then configure the new system
+        app.system = system
+        app.system.configure()
+
+        // Start
+        system.start(networksNeeded: [])
```
michael-brd: I thought start was removed, does this build?

EBGToo: Drat.

EBGToo: Fixed; ran and Crypto unit tests pass (Core test fail on a ETH P2P issue).

PR #116, opened 2019-07-29
Fix eth testnet network name

acorscadden:
A blockchain's network string is used to generate ethquery urls. Currently, the urls are: `/ethq/testnet/query`, but they should be `/ethq/ropsten/query`.


EBGToo:
This is not the proper place to fix the issue.  Fixed with CORE-450

PR #120, opened 2019-08-01
CORE-436: Define BRCryptoKey in C; Respect BRAddressParams

EBGToo:


Swift/BRCrypto/common/BRCryptoKey.swift
```diff
@@ -192,93 +148,45 @@ public final class Key {
 
         var secret = secret
         defer { secret = UInt256() }
+        // Serialization - Public Key
 
         BRKeySetSecret (&core, &secret, 1)
```
michael-brd: Could this be moved to C as well? Selfish goal is to have no non-crypto calls in Java (and Swift).

Swift/BRCrypto/common/BRCryptoSigner.swift
```diff
@@ -7,6 +7,7 @@
 //
 
 import Foundation
+import BRCryptoC
```
michael-brd: Is it a future goal to have this in C as well?

crypto/BRCryptoKey.c
```diff
@@ -0,0 +1,278 @@
...
+//
+// Our serialization/deserialization works because we use the same address params.  If the above
+// questions are answered, the various cryptoKeyCreate*() functions will need to specify their
+// address params AND (AND AND) the serialization will need to WRITE OUT the specific params.
+//
+#define CRYPTO_ADDRESS_PARAMS  ((BRAddressParams) { \
+    3 + BITCOIN_PUBKEY_PREFIX,  \
+    3 + BITCOIN_SCRIPT_PREFIX,  \
+    3 + BITCOIN_PRIVKEY_PREFIX, \
+    "cry" \
```
michael-brd: "cry"?

crypto/BRCryptoKey.c
```diff
@@ -0,0 +1,278 @@
...
+extern size_t
+cryptoKeySerializePublic (BRCryptoKey key, /* ... */ uint8_t *data, size_t dataCount) {
+    size_t publicKeySize = BRKeyPubKey (&key->core, NULL, 0);
+    size_t serializeSize = 1 + publicKeySize;
+
+    if (NULL == data) return serializeSize;
+    if (dataCount < serializeSize) return 0;
+
+    data[0] = SERIALIZE_PUBLIC_IDENTIFIER;
+    BRKeyPubKey (&key->core, &data[1], dataCount);
```
michael-brd: dataCount - 1?

Swift/BRCrypto/common/BRCryptoKey.swift
EBGToo: Yes.

crypto/BRCryptoKey.c
```diff
@@ -0,0 +1,278 @@
...
+    size_t privateKeySize = BRKeyPrivKey (&key->core, NULL, 0, key->coreAddressParams);
+    size_t serializeSize  = 1 + privateKeySize;
+
+    if (NULL == data) return serializeSize;
+    if (dataCount < serializeSize) return 0;
+
+    data[0] = SERIALIZE_PRIVATE_IDENTIFIER;
+
+    // Make a string
+    char privKey[dataCount]; // includes 1 for '\0'
```
michael-brd: data is an input, output or both? Trying to understand why we copy it to privKey and then call BRKeyPrivKey

crypto/BRCryptoKey.c
EBGToo: This is an intentionally nonsensical (but workable) BRAddressParams definition.  For use when we don't have any other info.

Swift/BRCrypto/common/BRCryptoSigner.swift
EBGToo: Oh, maybe.  Everything in common/ that touchs bitcoin needs to be in C, it seems?

michael-brd: That would be ideal, from my POV, so that we don't have parallel implementations.

crypto/BRCryptoKey.c
EBGToo: The serialization code is no longer used, at this time.  Maybe best to remove it from the C.  Let me address some of Ehsan comments and see where it goes.

PR #121, opened 2019-08-01
CORE-445

EBGToo:
This is a large merge, it includes `master` into `crypto`.  The recent commits, labelled as CORE-248 (Voisine commits), CORE-436 (just reviewed and merged), and as CORE-445 are the interesting ones...


Java/Core/build.gradle
```diff
@@ -59,14 +59,19 @@ android {
         // Default: required
         release {
             minifyEnabled false
-            //proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
+
+            // Ensure that the -DNDEBUG compiler flag IS NOT set thereby ensuring that `assert(c)``
+            // evaluates `c` (for side-effects - but we've removed side-effects anyhow)
+            debuggable true
```
michael-brd: This doesn't seem right. Will the app debuggable override this?

EBGToo: Me commenting on build.gradle would be a guess, at best (event if GIT blame shows me).

PR #122, opened 2019-08-01
CORE-450: Use ethq/ropsten/query in JSON RPC requests.

EBGToo:
Needs Java update - are API ethq/testnet/proxy requests working?  It should be ethq/ropsten/proxy where 'ropsten' is an alternate name for 'testnet'.  We use 'testnet' throughout; the API needs 'ropsten' - thus I only changed the HTTP query path.

### DIFF
--- a/Swift/BRCrypto/common/BRBlockChainDB.swift
+++ b/Swift/BRCrypto/common/BRBlockChainDB.swift
@@ -1565,10 +1565,15 @@ public class BlockChainDB {
         }
     }
 
+    private func apiGetNetworkName (_ name: String) -> String {
+        let name = name.lowercased()
+        return name == "testnet" ? "ropsten" : name
+    }
+    
     internal func apiMakeRequestJSON (network: String,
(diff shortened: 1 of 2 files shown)


Swift/BRCrypto/common/BRBlockChainDB.swift
```diff
@@ -1565,10 +1565,15 @@ public class BlockChainDB {
         }
     }
 
+    private func apiGetNetworkName (_ name: String) -> String {
+        let name = name.lowercased()
```
michael-brd: The C code lowecases the name, but doing it here too has no impact.

We could do the testnet/ropsten substitution at the C layer as well, but I can see the pros and cons of that. What do you think?

EBGToo: Let's leave it here for now; we are expecting the whole JSON_RPC to disappear.  Admittedly, I didn't look/consider the C layer...

PR #124, opened 2019-08-02
CORE-439:  Create an interface for private key import

michael-brd:


Swift/BRCrypto/common/BRCryptoKey.swift
```diff
@@ -29,6 +37,21 @@ public final class Key {
...
+    /// Create `Key` from `string` using the passphrase to decrypt it. The string must be BIP38
+    /// format. Different crypto currencies have different implementations; this function will
+    /// look for a valid string using BITCOIN mainnet and BITCOIN testnet.
+    ///
+    /// - Parameter string
+    /// - Parameter passphrase
+    ///
+    /// - Returns: A Key if one exists
+    ///
+    static public func createFromString (asPrivate string: String, withPassphrase: String) -> Key? {
```
EBGToo: Given that BIP38 defines encrypting private keys, this function should be analogous to `createFromString:asPrivate` - and it is.

Swift/BRCrypto/BRCryptoWallet.swift
```diff
@@ -172,6 +187,12 @@ public final class Wallet: Equatable {
         }
     }
 
+    public func estimateFeeToImportFrom (key: Key,
+                                         fee: NetworkFee,
+                                         completion: @escaping (Result<TransferFeeBasis,FeeEstimationError>) -> Void) {
+        precondition (false)
```
EBGToo: We have JIRA issues for implementing these?

bitcoin/BRBIP38Key.c
```diff
@@ -181,7 +181,7 @@ int BRKeySetBIP38Key(BRKey *key, const char *bip38Key, const char *passphrase)
     
     BRKeySetSecret(key, &secret, flag & BIP38_COMPRESSED_FLAG);
     var_clean(&secret);
-    BRKeyLegacyAddr(key, address.s, sizeof(address), BIP38_ADDR_PARAMS);
+    BRKeyLegacyAddr(key, address.s, sizeof(address), params);
```
EBGToo: Aaron's definition of BIP38_ADDR_PARAMS is:
```
#define BIP38_ADDR_PARAMS      ((BRAddressParams) { BITCOIN_PUBKEY_PREFIX, 0, 0, NULL })
```
where only the first field is in common with `BITCOIN_ADDR_PARAMS`.  It seems only the first field is ever used in BIP38 contexts - do we know that for sure?  Is it known that BIP38 applies to testnet as well?  Given that only the first field is used and the mainnet/testnet differ by the first field, I'd expect BIP38 to apply to testnet as well (even though there is no reason to secure a testnet wallet).

EBGToo: We could remove `#define BIP38_ADDR_PARAMS ...` too?

crypto/BRCryptoKey.c
```diff
@@ -113,15 +126,34 @@ cryptoKeyFindAddressParams (const char *string) {
     return EMPTY_ADDRESS_PARAMS;
 }
 
+extern BRCryptoKey
+cryptoKeyCreateFromStringProtectedPrivate (const char *privateKey, const char * passphrase) {
+    if (!BRBIP38KeyIsValid (privateKey)) return NULL;
```
EBGToo: So with this function we create a `BRCryptoKey` based on BIP38.  Once returned, all 'BIP38 context' is lost - meaning that `BITCOIN_ADDRESS_PARAMS`will apply in any subsequent `BRCryptoKey` uses.  Is that a problem?

crypto/BRCryptoKey.c
```diff
@@ -167,21 +202,30 @@ cryptoKeyCreateForBIP32ApiAuth (const char *phrase, const char *words[]) {
...
 
-
     UInt512 seed;
     BRBIP39DeriveKey (seed.u8, phrase, NULL);
 
     BRKey core;
     BRBIP32BitIDKey (&core, &seed, sizeof(UInt512), index, uri);
 
-    return cryptoKeyCreateInternal(core, CRYPTO_ADDRESS_PARAMS);
+    BRCryptoKey result = cryptoKeyCreateInternal(core, CRYPTO_ADDRESS_PARAMS);
```
EBGToo: Thanks for all the correct cleanups.

bitcoin/BRBIP38Key.c
michael-brd: The BIP (https://github.com/bitcoin/bips/blob/master/bip-0038.mediawiki) makes no mention of mainnet/testnet. Other implementations support both so its not clear why we would restrict to mainnet.

Yes, BIP38_ADDR_PARAMS can and should be removed. I will amend.

crypto/BRCryptoKey.c
michael-brd: Based on iOS' usage, it does not appear to be.

They extend the `String` class with `isValidPrivateKey` and `isValidBip38Key` methods (the latter of which `cryptoKeyIsProtectedPrivate` is analogous to).

Beyond that, they have a `BRKey(bip38Key: address, passphrase: password)` method, which `cryptoKeyCreateFromStringProtectedPrivate (const char *privateKey, const char * passphrase)` is analogous to.

The address params associated with a key are used for serializing, at the moment. It doesn't look like that is used with the BIP38 logic. If that ends up being needed in the future, the serialization logic likely needs to be extended (i.e. support for specifying a password, maybe support for specifying a network, etc.).

Swift/BRCrypto/BRCryptoWallet.swift
michael-brd: Not at the moment. I can create now. Had been waiting for this to get in.

PR #132, opened 2019-08-07
CORE-465: Add support for signatures in raw format

ewerx:
- add interface for DER and and Raw signatures in BRCryptoSigner
- add BRKeySignRaw and DER to Raw/JOSE conversion for signatures in BRKey


EBGToo:
As this impacts `support/BRKey.*`, there needs to be a test case in `bitcoin/test.c` in `BRKeyTests()` near L1280.  The test there can be identical to the test in BRCryptoCommonTests.swift.

support/BRKey.c
```diff
@@ -364,7 +372,84 @@ size_t BRKeyLegacyAddr(BRKey *key, char *addr, size_t addrLen, BRAddressParams p
     return addrLen;
 }
 
-// signs md with key and writes signature to sig
+// convert signature from DER to raw encoding
+int BRSignatureDERToRaw (const uint8_t *derSig, size_t derSigLen, uint8_t *rawSig, size_t rawSigLen)
```
michael-brd: I think this function could be static, what do you think?

support/BRKey.c
```diff
@@ -364,7 +372,84 @@ size_t BRKeyLegacyAddr(BRKey *key, char *addr, size_t addrLen, BRAddressParams p
...
+    //Skip 0x30
+    int offset = 1;
+    if (derSig[1] == 0x81) {
+        //Skip sign
+        offset++;
+    }
+    
+    //Convert to unsigned. Should match DER length - offset
+    int encodedLength = derSig[offset++] & 0xff;
+    assert(encodedLength == derSigLen - offset);
```
michael-brd: I think we want to fail out (maybe in addition to an assert) since the assert may be a NOP depending on how this source is built.

support/BRKey.c
```diff
@@ -364,7 +372,84 @@ size_t BRKeyLegacyAddr(BRKey *key, char *addr, size_t addrLen, BRAddressParams p
     return addrLen;
 }
 
-// signs md with key and writes signature to sig
+// convert signature from DER to raw encoding
+int BRSignatureDERToRaw (const uint8_t *derSig, size_t derSigLen, uint8_t *rawSig, size_t rawSigLen)
+{
+    const int sigLen = 64;
```
michael-brd: I think we should avoid using signed types throughout here. I know the original Java used signed types, but unsigned types aren't readily available in Java. By sticking to unsigned types and verifying every bit of arithmetic for overflows, I think we will be better off.

ewerx:
After discussion with @voisine, I renamed the new functions `BRKeySignJOSE` and `BRKeyVerifyJOSE` and refactored them to use the existing `secp256k1_ecdsa_signature_serialize_compact` serialization function instead of the DER to JOSE conversion.
Added tests.
Updated CryptoSwift interface and test.

support/BRKey.c
```diff
@@ -53,6 +53,14 @@
 #pragma clang diagnostic pop
 #pragma GCC diagnostic pop
 
+#if !defined (MAX)
```
EBGToo: Are MAX and MIN needed now?

Swift/BRCrypto/common/BRCryptoSigner.swift
```diff
@@ -77,6 +78,15 @@ public enum CoreSigner: Signer {
 
                     BRKeySign (cryptoKeyGetCore(key), targetAddr, targetCount, digestUInt256)
                 }
+                
+            case .basicJOSE:
+                let targetCount = BRKeySignJOSE (cryptoKeyGetCore(key), nil, 0, digestUInt256)
+                if 0 == targetCount { /* error */ }
```
EBGToo: Do you have a sense if these `0 == targetCount` errors are possible?  If so, should the return be `Data?` (or Result<Data,SignerError>` (if `SignerError` is enumerable)?

support/BRKey.c
ewerx: nope, removed.

Swift/BRCrypto/common/BRCryptoSigner.swift
ewerx: I stepped through all the signing/serialization functions and they will always return 1. The only scenario where `BRKeySign` returns 0 is if the length passed in less than required and the target buffer is non-nil. I replaced the `targetCount == 0` error checks with asserts.

PR #140, opened 2019-08-08
CORE-311: Remove unnecessary transfer add that triggered assert

michael-brd:
Adding transfer directly in createTransfer routine caused ETH assert on NULL == transfer in BRCryptoWalletManagerClient's TRANSFER_EVENT_CREATED handler.

I believe the pattern we've established is that transfers are added in their event handler so restoring that pattern with this commit.

In my testing, transfer status events (created, submitted, included) are all propagated correctly.


EBGToo:
>...
> 
> I believe the pattern we've established is that transfers are added in their event handler so restoring that pattern with this commit.
> 
> ...

 That is the pattern we need to follow, I think.

PR #141, opened 2019-08-08
CORE-472: Release resources in BRWalletManagerFree

michael-brd:


bitcoin/BRWalletManager.c
```diff
@@ -638,9 +638,18 @@ BRWalletManagerNew (BRWalletManagerClient client,
 
 extern void
 BRWalletManagerFree (BRWalletManager manager) {
-    fileServiceRelease(manager->fileService);
-    BRPeerManagerFree(manager->peerManager);
-    BRWalletFree(manager->wallet);
+    BRPeerManagerDisconnect (manager->peerManager);
+    BRPeerManagerFree (manager->peerManager);
```
EBGToo: This is probably going to expose a race... which, if it occurs, we'll deal with later.

michael-brd: Agreed. We know as a whole that BRWalletManager needs some threading-related changes.

PR #146, opened 2019-08-09
CORE-480: Start the EWM event handler on `ewmStart()` not on `ewmConnect()`.

EBGToo:
CWM will call ewmCreate(...), set up `cwm` and `cwm->wallet`, and then call `ewmStart()` to generate EWM events.  It might be that CWM doesn't need to generate its own CWM events (in cwmCreate()) as the EWM events, once flowing, will generate CWM events(?).

### DIFF
--- a/ethereum/ewm/BREthereumEWM.c
+++ b/ethereum/ewm/BREthereumEWM.c
@@ -480,7 +480,8 @@ ewmCreate (BREthereumNetwork network,
     // Queue the CREATED event so that it is the first event delievered to the BREthereumClient
     ewmSignalEWMEvent (ewm, EWM_EVENT_CREATED, SUCCESS, NULL);
 
-    // Create a default ETH wallet; other wallets will be created 'on demand'
+    // Create a default ETH wallet; other wallets will be created 'on demand'.  This will signal
+    // a WALLET_EVENT_CREATED event.
     ewm->walletHoldingEther = walletCreate(ewm->account,
                                            ewm->network);
     ewmInsertWallet(ewm, ewm->walletHoldingEther);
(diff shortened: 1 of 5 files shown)


ethereum/ewm/BREthereumEWM.c
```diff
@@ -653,6 +663,27 @@ ewmDestroy (BREthereumEWM ewm) {
...
+
+    // Start the EWM thread
+    eventHandlerStart(ewm->handler);
+}
+
+extern void
+ewmStop (BREthereumEWM ewm) {
+    // Stop the alarm clock
+    alarmClockStop (alarmClock);
+
```
michael-brd: Devil's advocate (not sure myself if we want this), but should stop imply disconnect? Almost like how on Android the lifecycle is onCreate->onStart->onResume->onPause->onStop->onDestroy (so, ewmCreate->ewmStart->ewmConnect->ewmDisconnect->ewmStop->ewmDestroy)?

EBGToo: Maybe it is worse.... disconnect => stop (which it shouldn't); stop !=> disconnect (which it should).

ethereum/ewm/BREthereumEWM.c
```diff
@@ -653,6 +663,27 @@ ewmDestroy (BREthereumEWM ewm) {
     free (ewm);
 }
 
+/// MARK: - Start/Stop
+
+extern void
+ewmStart (BREthereumEWM ewm) {
+    // Start the alarm clock.
+    alarmClockStart(alarmClock);
```
michael-brd: This is the process-wide alarm clock, ya? Don't think we want to be doing this, as it has repercussions on any other handler using a timeout, if I am understanding correctly.

EBGToo: CORE-489

ethereum/ewm/BREthereumEWM.c
EBGToo: Updated: `ewmStop()` calls `ewmDisconnect()`.  

Added a number of TODOs - for example, noting that start() and stop() don't check on the current state (which they arguably should).

EBGToo:
Can we move this to 'approved' understanding that we've more work to do on the sequencing but, with merging, to allow the testing of ETH to continue?

PR #148, opened 2019-08-12
CORE-478: Extend BRCrypto{WalletManager, Wallet, Transfer} Tests.

EBGToo:
This does not completely cover the WalletManager, Wallet and Transfer interface - but, aside from 'submit', it is pretty good.  Would expect to add more tests as some existing bugs are fixed.

This is all in Swift; certain things must be; some others could be in C, at some point.

This does include a merge from CORE-480 (EWM Event Handlers) and we'd expect CORE-480 to be reworked in a future merge to 'crypto'


michael-brd:
Comments from the other review (of addition to ewmStart/ewmStop) apply here.

EBGToo:
> Comments from the other review (of addition to ewmStart/ewmStop) apply here.

Agreed.  I'm expecting the ewmStart/Stop issues will get resolved in the other JIRA/PR.  (The committed included here from CORE-480 won't be reverted when it is fixed, it will be superseded; not trying to backdoor the CORE-480 change.)

EBGToo:
Removed the CORE-480 merge (to clarify the changes in this PR).

PR #152, opened 2019-08-14
CORE-392: Refactor connection/sync/submission implementation for BTC 

michael-brd:


bitcoin/BRWalletManager.c
```diff
@@ -1038,134 +733,188 @@ _BRWalletManagerTxDeleted (void *info,
...
+        }
 
-static void
-_BRWalletManagerSavePeers  (void *info,
-                            int replace,
-                            OwnershipKept const BRPeer *peers,
-                            size_t count) {
-    BRWalletManager manager = (BRWalletManager) info;
+        case SYNC_MANAGER_SET_PEERS: {
+            fileServiceClear(bwm->fileService, fileServiceTypePeers);
```
EBGToo: Does the `fileService` need to remain in `BWM`?  Everything the file service saves (at least currently) is related to syncing and thus maybe it is appropriately in the sync manager?

bitcoin/BRSyncManager.c
```diff
@@ -0,0 +1,1549 @@
...
+#define MAX(a,b) (((a)>(b))?(a):(b))
+#endif
+
+#if !defined (MIN)
+#define MIN(a,b) (((a)<(b))?(a):(b))
+#endif
+
+struct BRSyncManagerStruct {
+    BRSyncMode mode;
+};
```
EBGToo: Certain other common things can already be moved into here?  `lock` and `isConnected` at least?

bitcoin/BRSyncManager.c
```diff
@@ -0,0 +1,1549 @@
...
+                                       blockHeight,
+                                       blocks,
+                                       blocksCount,
+                                       peers,
+                                       peersCount);
+        default:
+        assert (0);
+        return NULL;
+    }
+}
```
EBGToo: Can you sketch how this is going to work for the other modes?  [Other modes are arguably not in this first release; but at least SYNC_MODE_BRD_WITH_P2P_SEND is coming along right fast.]

bitcoin/BRSyncManager.c
```diff
@@ -0,0 +1,1549 @@
...
+    BRPeerManagerPublishTx (manager->peerManager,
+                            transaction,
+                            info,
+                            _BRPeerSyncManagerTxPublished);
+}
+
+static void
+BRPeerSyncManagerTickTock(BRPeerSyncManager manager) {
+    double progress = BRPeerManagerSyncProgress (manager->peerManager, 0);
+    uint8_t needSyncEvent = progress < 1.0;
```
EBGToo: `0.0 < progress && progress < 1.0`?  Symmetrically.

bitcoin/BRSyncManager.c
```diff
@@ -0,0 +1,1549 @@
...
+    BRPeerSyncManager manager = (BRPeerSyncManager) info;
+    manager->eventCallback (manager->eventContext,
+                            BRPeerSyncManagerAsSyncManager (manager),
+                            (BRSyncManagerEvent) {
+                                replace ? SYNC_MANAGER_SET_PEERS : SYNC_MANAGER_ADD_PEERS,
+                                {
+                                    .peers = { (BRPeer *) peers, count }
+                                }
+                            });
+}
```
EBGToo: These Bitcoin callbacks (here and others) are happening in the Bitcoin thread and the `eventCallback` is a) guaranteed not to save event state (like `peers` above) and generally b) not to block excessively?

bitcoin/BRSyncManager.c
michael-brd: Came down to preference. I can move but has no practical impact. I personally like looking at the struct and seeing all its fields in one place rather than bouncing around.

bitcoin/BRSyncManager.c
michael-brd: We would have a new SyncManager implementation. How it would work internally we'd determine at that time. Trying to plan for that now seems like a guess at best. We can (and likely will) refactor at that time.

bitcoin/BRSyncManager.c
michael-brd: I had that, took it out, thought they might want the 0% case announced. Can put back in.

bitcoin/BRSyncManager.c
michael-brd: I preserved the existing behaviour here because you had expressed concerns about persistence in the past. The handling of this event could be moved to the event thread but I erred on the side of existing behaviour.

bitcoin/BRSyncManager.h
```diff
@@ -0,0 +1,217 @@
...
+        } syncStopped;
+        struct {
+            uint64_t value;
+        } blockHeightUpdated;
+        struct {
+            BRTransaction *transaction;
+            int error;
+        } submitted;
+    } u;
+} BRSyncManagerEvent;
```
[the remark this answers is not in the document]
michael-brd: The complexity is high already; I didn't focus on a GEN version. Main focus was correctness. If this proves to be the right implementation, we can then factor out the client version.

bitcoin/BRSyncManager.h
```diff
@@ -0,0 +1,217 @@
...
+BRSyncManagerNewForP2P(BRSyncManagerEventContext eventContext,
+                       BRSyncManagerEventCallback eventCallback,
+                       OwnershipKept const BRChainParams *params,
+                       OwnershipKept BRWallet *wallet,
+                       uint32_t earliestKeyTime,
+                       uint64_t blockHeight,
+                       OwnershipKept BRMerkleBlock *blocks[],
+                       size_t blocksCount,
+                       OwnershipKept const BRPeer peers[],
+                       size_t peersCount);
```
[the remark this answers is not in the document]
michael-brd: They do not need to be, no. I can make them static.

bitcoin/BRSyncManager.h
```diff
@@ -0,0 +1,217 @@
...
+
+extern void
+BRSyncManagerFree(BRSyncManager manager);
+
+extern void
+BRSyncManagerConnect(BRSyncManager manager);
+
+extern void
+BRSyncManagerDisconnect(BRSyncManager manager);
+
```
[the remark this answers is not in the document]
michael-brd: In past conversations, it sounded like you weren't keen on yet another event handler. So, used this approach for periodic events (BDB height/txn polling, P2P progress).

bitcoin/BRWalletManager.c
michael-brd: `fileService` is used by the BRWallet callbacks as well. Personally, I think `fileService` should be wrapped in a manager (BRFileManager?) and this could potentially be used internally `BRWalletManager` and `BRSyncManager`.

I'm personally of mixed feelings on this. This thought occurred to me too but I chose to keep it where it is now given the BRWallet callbacks.

bitcoin/BRWalletManager.h
```diff
@@ -207,18 +207,22 @@ typedef enum {
...
     BITCOIN_WALLET_MANAGER_SYNC_STOPPED,
     BITCOIN_WALLET_MANAGER_BLOCK_HEIGHT_UPDATED
 } BRWalletManagerEventType;
 
 typedef struct {
     BRWalletManagerEventType type;
     union {
+        struct {
+            uint32_t percentComplete;
+        } syncProgress;
```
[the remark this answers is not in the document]
michael-brd: I can't imagine anyone cares about 49.5% vs 50%.

bitcoin/BRWalletManagerPrivate.h
```diff
@@ -54,14 +54,9 @@ struct BRWalletManagerStruct {
...
-     * The BlockHeight is the largest block number seen
+     * The forkId of the chain parameters associated with the wallet
      */
-    uint32_t blockHeight;
-
-    /**
-     * An identiifer for a BRD Request
-     */
-    unsigned int requestId;
+    uint8_t forkId;
```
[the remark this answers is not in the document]
michael-brd: Where can I get the chainParams? The wallet doesn't store them. Are you saying store that, instead of the forkId, to get the forkId?

crypto/BRCryptoWallet.c
```diff
@@ -502,7 +502,7 @@ cryptoWalletCreateTransfer (BRCryptoWallet  wallet,
             uint64_t feePerKb      = cryptoFeeBasisAsBTC(estimatedFeeBasis);
 
             BRWalletSetFeePerKb (wid, feePerKb);
-            BRTransaction *tid = BRWalletManagerCreateTransaction (bwm, wid, value, addr);
+            BRTransaction *tid = BRWalletManagerCreateTransaction (bwm, value, addr);
```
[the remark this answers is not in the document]
michael-brd: `BRWalletCreateTransaction` and `BRWalletSubmitTransaction` had different approaches (one took wallet, one didn't). I made them both omit it. Looks like I guessed wrong! I've made them both require it.

crypto/BRCryptoWalletManager.c
```diff
@@ -138,6 +138,8 @@ cryptoWalletManagerCreate (BRCryptoCWMListener listener,
                                              cwmPath,
                                              cryptoNetworkGetHeight(network));
 
+            BRWalletManagerStart (cwm->u.btc);
+
```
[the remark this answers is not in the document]
michael-brd: No, because we call back with the wallet inline in `BRWalletManagerNew`. When that changes, we will need to behave like ETH. See the comment in `BRWalletManagerNew` for details.

bitcoin/BRSyncManager.c
EBGToo: But that is what I'm asking - just for a flavor of how it might work.  It seems to me that the other modes will have a bit of `ForClient()` and a bit of `ForP2P`.  It is almost like the wallet manager always needs `ForClient()` and then for BTC and ETH has an optional `ForP2P`.

crypto/BRCryptoWalletManagerClient.c
```diff
@@ -106,7 +106,7 @@ cwmGetBlockNumberAsBTC (BRWalletManagerClientContext context,
 static void
 cwmGetTransactionsAsBTC (BRWalletManagerClientContext context,
                          BRWalletManager manager,
-                         const char **addresses,
+                         OwnershipGiven const char **addresses,
```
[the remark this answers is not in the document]
michael-brd: It did not, no. Would you like the `const` stripped?

bitcoin/BRSyncManager.c
EBGToo: Some of the events are handled with a `signal` and some are handled directly (like file service ones).  I think the file service ones must be handled directly - because the peers, blocks, transactions are all stack allocated in the Bitcoin code.

crypto/BRCryptoWalletManager.c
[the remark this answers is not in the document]
EBGToo: Okay, I thought this change would include that change - but didn't confirm in the code.

bitcoin/BRSyncManager.c
michael-brd: They all have copy routines (i.e. peers and merkleblocks) so they could be moved to the event handler if you'd like.

crypto/BRCryptoWalletManagerClient.c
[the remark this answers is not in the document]
EBGToo: not a big deal.  If the static analyzer shows it (CORE-250), we can address then.

bitcoin/BRSyncManager.h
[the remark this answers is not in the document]
EBGToo: I think it is fine; it was a comment for clarification.

bitcoin/BRWalletManagerPrivate.h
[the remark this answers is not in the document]
EBGToo: The BRPeerManager stores them.  The `forkId` can only matter to the `ForP2P()` sync manager?  [Not a big deal; just surprised to see it.]

bitcoin/BRSyncManager.c
EBGToo: How about a comment in the handler code for those cases?  To the effect of "Peers may be stack allocated by the caller; we'll process inline". Similarly for blocks, transactions.

michael-brd: Ok. I had put a `OwnershipKept` on `_BRWalletManagerSyncEvent` to make that clear, but I can add a comment to be more explicit.

bitcoin/BRWalletManagerPrivate.h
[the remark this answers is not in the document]
michael-brd: No, the forkId is used for signing, which is outside of the `BRSyncManager`.

bitcoin/BRWalletManagerEvent.c
```diff
@@ -13,6 +13,102 @@
...
+bwmSignalTransactionEventDispatcher (BREventHandler ignore,
+                                     BRWalletManagerTransactionEvent *event) {
+    bwmHandleTransactionEvent(event->manager, event->wallet, event->transaction, event->event);
+}
+
+static BREventType bwmTransactionEventType = {
+    "BTC: Transaction Event",
+    sizeof (BRWalletManagerTransactionEvent),
+    (BREventDispatcher) bwmSignalTransactionEventDispatcher
+};
```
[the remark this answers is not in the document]
michael-brd: No, the pattern with ETH (and with BTC once we have proper transaction ownership) is that all transactions are owned by the Wallet (or in BTC's case, likely the WalletManager). So, we don't destroy transactions in events.

PR #153, opened 2019-08-14
CORE-483: Additional Key and Support Tests

EBGToo:


Swift/BRCryptoTests/BRCryptoCommonTests.swift
```diff
@@ -98,16 +99,44 @@ class BRCryptoCommonTests: XCTestCase {
         //
         // Phrase
         //
-        k = Key.createFrom (phrase: "nothing", words: BRCryptoAccountTests.words)
-        XCTAssertNil(k)
         k = Key.createFrom (phrase: "ginger settle marine tissue robot crane night number ramp coast roast critic", words: BRCryptoAccountTests.words)
         XCTAssertNotNil(k)
+        k = Key.createFrom (phrase: "ginger settle marine tissue robot crane night number ramp coast roast critic", words: nil)
```
michael-brd: Shouldn't we make `word` a non-optional parameter?

Swift/BRCryptoTests/BRCryptoCommonTests.swift
```diff
@@ -98,16 +99,44 @@ class BRCryptoCommonTests: XCTestCase {
...
+        XCTAssertNil(k)
+        k = Key.createFrom (phrase: "not-a-chance", words: BRCryptoAccountTests.words)
+        XCTAssertNil(k)
 
         // Pigeon
+        s = "5Kb8kLf9zgWQnogidDA76MzPL6TsZZY36hWXMssSzNydYXYB9KF"
+        k = Key.createFromString(asPrivate: s)
+        n = "nonce".data(using: .utf8)
+        l = Key.createForPigeonFrom(key: k, nonce: n)
```
michael-brd: Do we want to assert on this being not nil?

Swift/BRCryptoTests/BRCryptoCommonTests.swift
```diff
@@ -98,16 +99,44 @@ class BRCryptoCommonTests: XCTestCase {
...
 
         // BIP32ApiAuth
+        k = Key.createForBIP32ApiAuth (phrase: "ginger settle marine tissue robot crane night number ramp coast roast critic", words: BRCryptoAccountTests.words)
+        XCTAssertNotNil(k)
+        k = Key.createForBIP32ApiAuth (phrase: "ginger settle marine tissue robot crane night number ramp coast roast critic", words: nil)
```
michael-brd: Same comment re: words being non-optional.

Swift/BRCryptoTests/BRCryptoCommonTests.swift
```diff
@@ -98,16 +99,44 @@ class BRCryptoCommonTests: XCTestCase {
...
         // BIP32BitID
+        k = Key.createForBIP32BitID (phrase: "ginger settle marine tissue robot crane night number ramp coast roast critic",
+                                     index: 2,
+                                     uri: "some uri",
+                                     words: BRCryptoAccountTests.words)
+        XCTAssertNotNil(k)
+        k = Key.createForBIP32BitID (phrase: "ginger settle marine tissue robot crane night number ramp coast roast critic",
+                                     index: 2,
+                                     uri: "some uri",
+                                     words: nil)
```
michael-brd: Same comment re: words being non-optional.

Swift/BRCryptoTests/BRCryptoCommonTests.swift
EBGToo: The idea was that you could install the words appropriate for your locale and then not worry about it again:
```
public final class Key {
    static public var wordList: [String]?
    // ...
    static public func createFrom (phrase: String, words: [String]? = wordList) -> Key? {
       // ...
```
Until the static `wordList` is assigned it will be `nil` and thus the optional argument, such as in `createFrom(phrase:words)`, must accept `nil`

EBGToo: So it is a) provide `words`, b) assign `wordList` or c) handle `nil` (owing to no words or a corrupt phrase).

Swift/BRCryptoTests/BRCryptoCommonTests.swift
EBGToo: Added a check.

PR #156, opened 2019-08-15
CORE-490: Add tests for BTC connect/sync events

michael-brd:


bitcoin/testBwm.c
```diff
@@ -0,0 +1,1063 @@
...
+BRRunTestWalletManagerSyncTestSetup (BRRunTestWalletManagerSyncState *state, uint64_t blockHeight) {
+    state->blockHeight = blockHeight;
+    array_new (state->managerEvents, 1);
+    array_new (state->walletEvents, 1);
+    array_new (state->transactionEvents, 1);
+    pthread_mutex_init (&state->lock, NULL);
+}
+
+static int
+BRRunTestWalletManagerSyncTestVerification (BRRunTestWalletManagerSyncState *state) {
```
EBGToo: This is checking that that events that did occur occurred in an allowed order but not that any particular event occurred?  Thus, you could have success with an event sequence of [WALLET_MANAGER_CREATED] only - w/o ever seeing WALLET_MANAGER_WALLET_ADDED?

bitcoin/testBwm.c
```diff
@@ -0,0 +1,1063 @@
...
+        case BITCOIN_WALLET_FEE_ESTIMATED:
+        return "BITCOIN_WALLET_FEE_ESTIMATED";
+        case BITCOIN_WALLET_DELETED:
+        return "BITCOIN_WALLET_DELETED";
+    }
+    return "<BITCOIN_WALLET_TYPE_UNKNOWN>";
+}
+
+static int
+_isValidWalletManagerEventPair (BRWalletManagerEvent *e1, BRWalletManagerEvent *e2) {
```
EBGToo: From a documentation viewpoint, this is good stuff.

EBGToo: Is it worth putting this in Crypto C - essentially right next to the `enum`?

Java/CoreNative/src/androidTest/java/com/breadwallet/corenative/CryptoLibraryAIT.java
```diff
@@ -26,6 +27,7 @@ public void setup() {
         // this is a compromised testnet paperkey
         paperKey = "ginger settle marine tissue robot crane night number ramp coast roast critic";
         coreDataDir = new File (context.getFilesDir(), "corenative");
+        epoch = 1504704195;
```
EBGToo: Time to move to `Date`?

bitcoin/testBwm.c
```diff
@@ -0,0 +1,1063 @@
...
+        return "BITCOIN_WALLET_TRANSACTION_SUBMITTED";
+        case BITCOIN_WALLET_FEE_PER_KB_UPDATED:
+        return "BITCOIN_WALLET_FEE_PER_KB_UPDATED";
+        case BITCOIN_WALLET_FEE_ESTIMATED:
+        return "BITCOIN_WALLET_FEE_ESTIMATED";
+        case BITCOIN_WALLET_DELETED:
+        return "BITCOIN_WALLET_DELETED";
+    }
+    return "<BITCOIN_WALLET_TYPE_UNKNOWN>";
+}
```
EBGToo: Maybe this in Crypto C as well?  The Swift has this:
```
extension BRWalletManagerEventType: CustomStringConvertible {
    public var description: String {
        switch self {
        case BITCOIN_WALLET_MANAGER_CREATED: return "Created"
        case BITCOIN_WALLET_MANAGER_CONNECTED: return "Connected"
        case BITCOIN_WALLET_MANAGER_DISCONNECTED: return "Disconnected"
        case BITCOIN_WALLET_MANAGER_SYNC_STARTED: return "Sync Started"
        case BITCOIN_WALLET_MANAGER_SYNC_PROGRESS: return "Sync Progress"
        case BITCOIN_WALLET_MANAGER_SYNC_STOPPED: return "Sync Stopped"
        case BITCOIN_WALLET_MANAGER_BLOCK_HEIGHT_UPDATED: return "Block Height Updated"
        default: return "<<unknown>>"
        }
    }
}
```
Ignoring the string differences, there should be duplicate strings. I think in ETH, we are pretty consistent with having `enum -> string` functions.

bitcoin/testBwm.c
michael-brd: For the String functions, I could see that. Having this valid pairs function in the library doesn't feel right.

EBGToo: Understood regarding pairs of `BRCryptoWalletManagerEventType`.  Perhaps one could argue have a pairs function for `BRCryptoWalletManagerState` does make sense as: 1) it essentially define the 'state chart transitions' and 2) supports validation of `CRYPTO_WALLET_MANAGER_EVENT_CHANGED` old and new state pairs.  But, no need to slow up this PR on that.

bitcoin/testBwm.c
EBGToo: ^edit with 'there should not be duplicate string...'

michael-brd: I don't follow; are you asking to have the Swift call this new `BRWalletEventTypeString` function once it is in C? I'm not sure why this is Swift at the moment; Swift should have no notion of BTC (or ETH) events.

bitcoin/testBwm.c
michael-brd: That is correct. I can add to the routine to confirm that WALLET_MANAGER_CREATED was received, and WALLET_MANAGER_WALLET_ADDED thereafter.

bitcoin/testBwm.c
EBGToo: Alas, seems the BTC and ETH 'type -> string' functions are hold overs from when System dealt with BTC and ETH directly, instead of CRYPTO as is now the case.  The System callback logging does do a 'type -> string' for `BRCryptoWalletManagerEventType` and others.

Java/CoreNative/src/androidTest/java/com/breadwallet/corenative/CryptoLibraryAIT.java
michael-brd:
```
        Calendar calendar = Calendar.getInstance(TimeZone.getTimeZone("UTC"));
        calendar.set(2017, /* September */ 8, 7);
        epoch = (int) TimeUnit.MILLISECONDS.toSeconds(calendar.getTimeInMillis());
```
The joys of Java.

bitcoin/testBwm.c
michael-brd: I added the BTC ones to the Crypto C. Of note, ETH does not have equivalent C functions.

PR #157, opened 2019-08-15
CORE-305: Rework Ethereum Events

EBGToo:
(Done in support of CORE-486 and CORE-491)

Ethereum event interfaces passed `(..., event (as an enum), status, error )`.  Changed to be 
```
typedef struct {
   type,
   status,
   union { /* event specific data */} u;
   error
} event;
```
This structured approach allows one to pass values back in an event.

### DIFF
--- a/crypto/BRCryptoWalletManagerClient.c
+++ b/crypto/BRCryptoWalletManagerClient.c
@@ -22,6 +22,9 @@
 //  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 //  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 //  THE SOFTWARE.
+//
+#include <math.h>  // round()
+
 #include "BRCryptoBase.h"
 #include "BRCryptoStatus.h"
 #include "BRCryptoPrivate.h"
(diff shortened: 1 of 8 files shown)


crypto/BRCryptoWalletManagerClient.c
```diff
@@ -851,41 +874,41 @@ cwmWalletManagerEventAsETH (BREthereumClientContext context,
...
         }
 
-        case EWM_EVENT_SYNC_STARTED:
-            cwmEvent = (BRCryptoWalletManagerEvent) {
-                CRYPTO_WALLET_MANAGER_EVENT_SYNC_STARTED
-            };
-            cwm->listener.walletManagerEventCallback (cwm->listener.context,
-                                                      cryptoWalletManagerTake (cwm),
-                                                      cwmEvent);
+        case EWM_EVENT_CHANGED:
```
michael-brd: Inferring this seems tricky. Interested in knowing what the issue was with the STARTED/STOPPED events.

crypto/BRCryptoWalletManagerClient.c
```diff
@@ -851,41 +874,41 @@ cwmWalletManagerEventAsETH (BREthereumClientContext context,
...
-            cwm->listener.walletManagerEventCallback (cwm->listener.context,
-                                                      cryptoWalletManagerTake (cwm),
-                                                      cwmEvent);
-
+        case EWM_EVENT_SYNC_PROGRESS:
             cwmEvent = (BRCryptoWalletManagerEvent) {
-                CRYPTO_WALLET_MANAGER_EVENT_CHANGED,
-                { .state = { cwm->state, CRYPTO_WALLET_MANAGER_STATE_CONNECTED }}
+                CRYPTO_WALLET_MANAGER_EVENT_SYNC_CONTINUES,
+                { .sync = { round (event.u.syncProgress.percent) }}
```
michael-brd: I shed a tear for the lost precision.

crypto/BRCryptoWalletManagerClient.c
EBGToo: (Actually, I was inclined to take it further - to sync started/stopped.)  The event is a change of state - the action is to do what you please based on a) the state you are exiting and b) the state you are entering.

ethereum/ewm/BREthereumClient.h
```diff
@@ -291,42 +291,46 @@ extern "C" {
...
         BREthereumWalletEventType type;
+        BREthereumStatus status;
         union {
             struct {
                 BREthereumCookie cookie;
                 BREthereumGas gasEstimate;
                 BREthereumGasPrice gasPrice;
             } feeEstimate;
         } u;
+        char errorDescription[16];
```
michael-brd: Is the status code not enough? What is captured here?

EBGToo: That `errorDescription` is filled in a single case where a transfer has failed.  It will have a string like 'gas limit exceeded' or 'gas price too low'.    Maybe the string could be avoided if `BREthereumStatus` listed up the failure reasons - which actually it might be able to do:
```
typedef enum {
    SUCCESS,
    // ...

    // Transfer
    ERROR_TRANSACTION_HASH_MISMATCH,
    ERROR_TRANSACTION_SUBMISSION,

    // ...
} BREthereumStatus;

crypto/BRCryptoWalletManagerClient.c
EBGToo: We could do with an addition to `support/<the-most-basis>.h` of `typedef <a number> BRSyncProgressPercentage` - given that we have X (X >=3) current representations.

ethereum/ewm/BREthereumClient.h
michael-brd: Yeah. For C, I"m certainly a fan of the single, massive error code type. It's not great but I haven't seen a better pattern in C.

EBGToo: VxWorks, I think IOS, have a concept of encoding a pair of {<module>, <module-error>} into an integer as an error code.  module is a global namespace (or could be declared/assigned dynamically).

PR #159, opened 2019-08-16
CORE-491: Announce EWM connected/disconnected/syncing events.

EBGToo:


ethereum/ewm/BREthereumEWM.c
```diff
@@ -727,6 +742,14 @@ ewmConnect(BREthereumEWM ewm) {
     }
 
     ewmUnlock (ewm);
+
```
michael-brd: With the locking scheme you have here, you have a window where if you have two racing threads calling `ewmConnect` and `ewmDisconnect`, the events could be queued in an order that the client doesn't expect (i.e. receives a DISCONNECTED->CONNECTED without first seeing CREATED/DISCONNECTED->CONNECTED)

This is what I ran into on the BTC front (and prompted the multithreaded tests.

EBGToo: Given that they are calls to `ewmSignal*()` which has no blocking implications, you think it is resolved by simply moving the calls into the lock region in the connect/disconnect functions?  Looks to me that that resolved it.  [The periodicDispatcher doesn't have an explicit lock/unlock but the eventHandler shares `ewm->lock` so that is okay].

michael-brd: Yes, that was what I had to do for BTC. Should work here.

PR #160, opened 2019-08-16
CORE-486: Test Event Sequences

EBGToo:
Not all tests pass...

Includes small changes to Swift interfaces (like to add 'Equatable'); generates a WalletManagerEvent.create event.  Otherwise test infrastructure and tests.


Swift/BRCrypto/BRCryptoTransfer.swift
```diff
@@ -236,12 +235,18 @@ public class TransferFeeBasis {
     deinit {
         cryptoFeeBasisGive (core)
     }
+
+    public static func == (lhs: TransferFeeBasis, rhs: TransferFeeBasis) -> Bool {
```
michael-brd: What about a C level `cryptoFeeBasisIsIdentical()`?

EBGToo: Added this (and associated ETH version).

PR #161, opened 2019-08-19
CORE-492: Fix Core module JNI to allow Core Crypto to build again

michael-brd:


Java/Core/src/main/cpp/jni/ethereum/com_breadwallet_core_ethereum_BREthereumEWM.c
```diff
@@ -1796,21 +1786,19 @@ clientGetNonce(BREthereumClientContext context,
...
                       BREthereumEWM ewm,
-                      BREthereumEWMEvent event,
-                      BREthereumStatus status,
-                      const char *errorDescription) {
+                      BREthereumEWMEvent event) {
     JNIEnv *env = getEnv();
     if (NULL == env) return;
 
-    jstring errorDescriptionString = (NULL == errorDescription
+    jstring errorDescriptionString = ('\0' == event.errorDescription[0]
```
EBGToo: I guess this is the safest, most direct, approach.  Can't meaningfully compare SUCCESS and event.type.

michael-brd: I sort of assumed you were setting the description to an empty string but is that not the case? This could be problematic if not. Should this be a check on if `SUCCESS == event.status`?

EBGToo: I don't think `errorDescription` is set consistently in the cases where the status is not `SUCCESS`.

PR #162, opened 2019-08-20
CORE-493: Allow dynamic mode changes on BRWalletManager

michael-brd:


michael-brd:
The `testBwm.c` file has some TODO statements. They unearthed some disparities between modes of relatively low priority consequence. Specifically, in P2P mode, if you call `BRPeerManagerConnect`, whilst connnected, it does a disconnect and re-connect. The current API behaviour is not the same as it simply stays connected.

I will create a JIRA issue to track addressing, one way or the other, and updating the tests.

bitcoin/BRWalletManager.h
```diff
@@ -210,6 +210,9 @@ typedef void
 extern const char *
 BRWalletEventTypeString (BRWalletEventType t);
 
+extern int
+BRWalletEventTypeIsValidPair (BRWalletEventType t1, BRWalletEventType t2);
+
```
EBGToo: Thanks.

bitcoin/BRWalletManager.c
```diff
@@ -690,17 +699,79 @@ BRWalletManagerGetWallet (BRWalletManager manager) {
...
+        // get the currently known block height
+        uint64_t blockHeight = BRSyncManagerGetBlockHeight (manager->syncManager);
+
+        // kill the sync manager to prevent any additional callbacks from occuring
+        BRSyncManagerDisconnect (manager->syncManager);
+        BRSyncManagerFree (manager->syncManager);
+
+        // load up the blocks/peers saved to disk
+        BRArrayOf(BRMerkleBlock*) blocks = initialBlocksLoad(manager);
+        BRArrayOf(BRPeer) peers = initialPeersLoad(manager);
```
EBGToo: These are surely written to the persistent store by now?

bitcoin/BRWalletManager.c
```diff
@@ -690,17 +699,79 @@ BRWalletManagerGetWallet (BRWalletManager manager) {
...
+                                                        },
+                                                        manager->chainParams,
+                                                        manager->wallet,
+                                                        manager->earliestKeyTime,
+                                                        blockHeight,
+                                                        blocks, array_count (blocks),
+                                                        peers, array_count (peers));
+    }
+    pthread_mutex_unlock (&manager->lock);
+}
```
EBGToo: When moving to P2P mode this `BRSyncManagerNewForMode` might induce a full sync (from some block before `earliestKeyTime`).  BRPeerManager will examine `blocks` and if not consistent it will, I believe, discard the blocks and start anew.  Thus unless the API mode gets the blocks in place to convince/fool BRPeerManager there will, I believe, be a full sync.

bitcoin/BRWalletManager.c
```diff
@@ -680,6 +688,7 @@ BRWalletManagerStart (BRWalletManager manager) {
 
 extern void
 BRWalletManagerStop (BRWalletManager manager) {
+    BRSyncManagerDisconnect (manager->syncManager);
```
EBGToo: Not a call to `BRWalletManagerDisconnect()`?  [If disconnecting involves more than just `BRSyncManagerDisconnect()`, you'd then be duplicating it.]

EBGToo: There is no `bwm->state` thingy to track?

bitcoin/BRWalletManager.c
michael-brd: As best can be reasoned, yes.

bitcoin/BRWalletManager.c
michael-brd: Correct. Is this an issue?

bitcoin/BRWalletManager.c
michael-brd: No `bwm->state`. I'm not a fan of each layer tracking information that is more accurate in a layer below it.

EBGToo: ^^^?

bitcoin/BRWalletManager.c
EBGToo: I think the expectation would be the User can use API mode to do a fast initial sync and then switch to P2P mode.  I don't think they'd expect a full sync at that time.

Once can a) get the blocks in place to convince/fool BRPeerManager or b) change the `earliestKeyTime` to be the time of the last block.

michael-brd: My understanding was that the initial use case was for P2P opting in to API, not a toggle back and forth. I'd put forth that we move ahead with this PR and track this as a separate issue as something like getting blocks will have a larger architectural impact.

PR #163, opened 2019-08-21
CORE-282: BRCryptoWalletMigrator

EBGToo:
Note that the 'Random Cleanup' has unrelated changes (in BRCryptoAmount.swfit, BRCryptoTransfer.swift, BRCryptoCommonTests.swift, and BRCryptoTransferTests.swift)

Includes changes to BRFileService to add a fileServiceCreateFromTypeSpecfications() - so that BRWalletManagerNew() and BRWalletManagerCreateFileService() can produce the same BRFileService for BTC.

### DIFF
--- a/Swift/BRCrypto/BRCryptoSystem.swift
+++ b/Swift/BRCrypto/BRCryptoSystem.swift
@@ -1166,33 +1166,68 @@ extension System {
     }
 }
 
+/// Support for Persistent Storage Migration.
+///
+/// Allow prior App version to migrate their SQLite database representations of BTC/BTC
+/// transations, blocks and peers into 'Generic Crypto' - where these entities are persistently
+/// stored in the file system (by BRFileSystem).
+///
 extension System {
-    public typealias TransactionBlob = (
-        bytes: [UInt8],
-        blockHeight: UInt32,
-        timestamp: UInt32) // time interval since unix epoch (including '0'
 
+    ///
+    /// A Blob of Transaction Data
+    ///
+    /// - btc:
+    ///
+    public enum TransactionBlob {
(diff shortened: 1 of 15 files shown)


Swift/BRCrypto/BRCryptoSystem.swift
```diff
@@ -1166,6 +1166,162 @@ extension System {
     }
 }
 
+extension System {
```
michael-brd: Is this really generic or is it BTC/BCH specific? The fields seem blockchain-specific, or are these all found in other blockchains?

EBGToo: It is BTC/BCH specific - because that is the only existing persistent data that needs to be migrated.  The code reads 'more generically' then is should.  Perhaps something like `TransferBlob` needs a BTC/BCH specific name or another approach altogether?

```
enum TransferBlob { 
    case btc ((bytes: [UInt8], blockHeight: UInt32, timestamp: UInt32))
}
```

michael-brd: That's an interesting idea; I kind of like that.

EBGToo: Let me see how it plays out.

EBGToo: Probably renaming `cryptoWalletMigratorHandleTransaction()` to `cryptoWalletMigratorHandleTransactionAsBTC()` too?

michael-brd: I guess so, and block? If this is more trouble than its worth, we can revisit later.

PR #164, opened 2019-08-21
CORE-437: In BRCryptoAddress, handle bitcoin cash addresses

EBGToo:
The behavior with testnet is pending.  (BCH testnet has unexpected behavior; a BTC testnet test case is needed).

### DIFF
--- a/crypto/BRCryptoAddress.c
+++ b/crypto/BRCryptoAddress.c
@@ -38,7 +38,7 @@ struct BRCryptoAddressRecord {
     BRCryptoBlockChainType type;
     union {
         struct {
-            int isBitcoinCashAddr;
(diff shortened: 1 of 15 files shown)


crypto/BRCryptoAddress.c
```diff
@@ -36,7 +37,10 @@ cryptoAddressRelease (BRCryptoAddress address);
 struct BRCryptoAddressRecord {
     BRCryptoBlockChainType type;
     union {
-        BRAddress btc;
+        struct {
+            int isBitcoinCashAddr;
```
michael-brd: Maybe BRCryptoBoolean?

EBGToo: Wow, that exposed a problem....

PR #167, opened 2019-08-23
CORE-494: Allow for ETH SyncMode Changes

EBGToo:
- Added ewm{Get,Update)Mode()
- Removed BREthereumMode, replaced with BRSyncMode
- Added ewmHandleSaveNodesEventDestroyer
- Updated the Swift Demo App significantly
- In ewm{Connect/Disconnect}() ensure EWM_EVENT_CHANGED is generated


crypto/BRCryptoWalletManager.c
```diff
@@ -327,6 +327,8 @@ cryptoWalletManagerSetMode (BRCryptoWalletManager cwm, BRSyncMode mode) {
             BRWalletManagerSetMode (cwm->u.btc, mode);
             break;
         case BLOCK_CHAIN_TYPE_ETH:
+            ewmUpdateMode (cwm->u.eth, mode);
```
michael-brd: ewmGetMode and ewmUpdateMode? You contrarian you... :)

PR #168, opened 2019-08-23
CORE-485: Add System.updateNetworkFees()

EBGToo:
@michael-brd What do you think of this?  Reasonable interface?  Workable for Android?


michael-brd:
Seems reasonable (please create a Java issue, in the unlikely event there isn't one already).

PR #170, opened 2019-08-23
CORE-453: Implement wallet sweeping interface

michael-brd:


Java/Crypto/src/main/java/com/breadwallet/crypto/Key.java
```diff
@@ -52,6 +52,10 @@ static boolean isProtectedPrivateKeyString(byte[] privatekeyUtf8) {
 
     byte[] encodeAsPublic();
 
+    byte[] encodeAsLegacyAddress();
+
+    byte[] encodeAsAddress();
+
```
EBGToo: What is the rationale for these?  On the surface, couldn't one take the result of `encodeAsPublic()` and then create a `BRCryptoAddress` (although such an interface might not exist yet).

Java/Crypto/src/main/java/com/breadwallet/crypto/WalletManager.java
```diff
@@ -9,12 +9,16 @@
...
+import com.breadwallet.crypto.errors.FeeEstimationError;
+import com.breadwallet.crypto.utility.CompletionHandler;
 import com.google.common.base.Optional;
 
 import java.util.List;
 
 public interface WalletManager {
 
+    void createSweeper(byte[] forPrivateKeyUt8, CompletionHandler<WalletSweeper, WalletSweeper.WalletSweeperError> completion);
+
```
EBGToo: Can't the first argument here be `Key` (w/ the expectation/check that it is a private key)?

crypto/BRCryptoWallet.h
```diff
@@ -203,6 +222,34 @@ extern "C" {
...
+    /// MARK: Wallet Sweeper
+
+    typedef enum {
+        CRYPTO_WALLET_SWEEPER_SUCCESS,
+        CRYPTO_WALLET_SWEEPER_INVALID_OPERATION,
+        CRYPTO_WALLET_SWEEPER_INVALID_HASH
+    } BRCryptoWalletSweeperStatus;
+
+    extern BRCryptoWalletSweeper
+    cryptoWalletSweeperCreateAsBtc (BRCryptoKey key, BRCryptoAddressScheme scheme);
```
EBGToo: Maybe given a `key` the sweeper does need more than one address?  That is, the UTXOs need to include LEGACY and SEGWIT public addresses?

bitcoin/BRWalletManager.h
```diff
@@ -326,6 +323,12 @@ BRWalletManagerCreateTransaction (BRWalletManager manager,
                                   const char *addr,
                                   uint64_t feePerKb);
 
+extern BRTransaction *
+BRWalletManagerCreateTransactionForSweep (BRWalletManager manager,
+                                          BRWallet *wallet,
+                                          BRWalletSweeper sweeper,
+                                          uint64_t feePerKb);
+
```
EBGToo: The `sweeper` will include the UTXOs?

crypto/BRCryptoWallet.h
```diff
@@ -203,6 +222,34 @@ extern "C" {
...
+
+    extern void
+    cryptoWalletSweeperRelease (BRCryptoWalletSweeper sweeper);
+
+    extern BRCryptoKey
+    cryptoWalletSweeperGetKey (BRCryptoWalletSweeper sweeper);
+
+    extern BRCryptoAmount
+    cryptoWalletSweeperGetBalance (BRCryptoWalletSweeper sweeper,
```
michael-brd: I don't like Unit being provided. I think instead of a "public" `cryptoWalletSweeperCreateAsBtc`, there should be a `cryptoWalletCreateSweeper` that will create one (and provided the correct unit).

bitcoin/BRWalletManager.h
EBGToo: Or will the BRWalletManager call out to BlockchainDB to accumulate them?

bitcoin/BRWalletManager.h
```diff
@@ -369,6 +385,30 @@ BRWalletManagerExtractFileServiceTypes (BRFileService fileService,
...
+//
+// Mark: Wallet Sweeper
+//
+
+typedef enum {
+    WALLET_SWEEPER_SUCCESS
+} BRWalletSweeperStatus;
+
+extern BRWalletSweeper // NULL on error
+BRWalletSweeperNew (BRKey *key, uint8_t isSegwit);
```
EBGToo: As mentioned previously, maybe the sweeper needs to sweep with both address schemes?

bitcoin/BRWalletManager.h
```diff
@@ -369,6 +385,30 @@ BRWalletManagerExtractFileServiceTypes (BRFileService fileService,
...
+
+extern void
+BRWalletSweeperFree (BRWalletSweeper sweeper);
+
+extern uint64_t
+BRWalletSweeperGetBalance (BRWalletSweeper sweeper);
```
EBGToo: Nice.

bitcoin/BRWalletManager.h
```diff
@@ -369,6 +385,30 @@ BRWalletManagerExtractFileServiceTypes (BRFileService fileService,
...
+
+extern BRWalletSweeperStatus
+BRWalletSweeperHandleTransactionOutput (BRWalletSweeper sweeper,
+                                        UInt256 hash,
+                                        uint32_t index,
+                                        OwnershipKept uint8_t *script,  size_t scriptLen,
+                                        uint64_t satoshis);
+
```
EBGToo: Ah, so this updates `sweeper` with the UTXO?

Java/Crypto/src/main/java/com/breadwallet/crypto/Key.java
michael-brd: I need access to the outputs of `BRKeyLegacyAddr` (and created one to wrap `BRKeyAddress` for parity). How I get access to that, I'm not fussed. But that API isn't being used anywhere at the moment in Crypto, so, did this.

Java/Crypto/src/main/java/com/breadwallet/crypto/WalletManager.java
michael-brd: Sure, easy enough to change.

crypto/BRCryptoWallet.h
michael-brd: The address scheme is so that I know what our receive address will be.

bitcoin/BRWalletManager.h
michael-brd: `BRWalletSweeperHandleTransactionOutput` is how the sweeper gets populated with UTXOs.

bitcoin/BRWalletManager.h
michael-brd: The existing iOS functionalty always uses legacy. The isSegwit is used for our received address. We don't store that information anywhere.

bitcoin/BRWalletManager.h
michael-brd: Correct. Modelled loosely on the migrator.

Java/Crypto/src/main/java/com/breadwallet/crypto/Key.java
michael-brd: I've got another idea here. How about the BRCryptoWalletSweeper has a cryptoWalletSweeperGetAddress function? Could be GetAddresses but for now its just one, so go with that, I'd say, to keep complexity down.

PR #180, opened 2019-08-29
CORE-454: Implement the interface for payment requests

michael-brd:


crypto/BRCryptoWallet.c
```diff
@@ -903,9 +903,11 @@ cryptoWalletSweeperValidateSupported (BRCryptoNetwork network,
         return CRYPTO_WALLET_SWEEPER_INVALID_ARGUMENTS;
     }
 
-    if (CRYPTO_FALSE == cryptoCurrencyIsIdentical (currency, cryptoWalletGetCurrency (wallet))) {
+    BRCryptoCurrency walletCurrency = cryptoWalletGetCurrency (wallet);
+    if (CRYPTO_FALSE == cryptoCurrencyIsIdentical (currency, walletCurrency)) {
         return CRYPTO_WALLET_SWEEPER_INVALID_ARGUMENTS;
     }
+    cryptoCurrencyGive (walletCurrency);
```
EBGToo: The 'give' is missed if the prior statement returns?

crypto/BRCryptoPayment.c
```diff
@@ -83,36 +85,40 @@ cryptoPaymentProtocolRequestBitPayBuilderCreate (BRCryptoNetwork cryptoNetwork,
...
+        builder->cryptoCurrency = cryptoCurrencyTake (cryptoCurrency);
+        builder->callbacks = callbacks;
 
-    if (network) {
-        builder->network = strdup (network);
-    }
+        builder->time = time;
+        builder->expires = expires;
+        builder->feePerByte = feePerByte;
+        array_new (builder->outputs, 10);
```
EBGToo: '10' was selected randomly, which I guess is a good as any.

crypto/BRCryptoPayment.c
```diff
@@ -234,46 +243,48 @@ static BRCryptoPaymentProtocolRequest
...
+            protoReq = calloc (1, sizeof (struct BRCryptoPaymentProtocolRequestRecord));
+            protoReq->ref = CRYPTO_REF_ASSIGN(cryptoPaymentProtocolRequestRelease);
+            protoReq->type = CRYPTO_PAYMENT_PROTOCOL_TYPE_BITPAY;
+
+            protoReq->cryptoNetwork = cryptoNetworkTake (builder->cryptoNetwork);
+            protoReq->cryptoCurrency = cryptoCurrencyTake (builder->cryptoCurrency);
+
+            BRCryptoUnit feeUnit = cryptoNetworkGetUnitAsBase (builder->cryptoNetwork, builder->cryptoCurrency);
+            BRCryptoAmount feeAmount = cryptoAmountCreateDouble (builder->feePerByte, feeUnit);
+            protoReq->u.btc.requiredFee = cryptoNetworkFeeCreate (0, feeAmount, feeUnit);
```
EBGToo: You are creating an essentially fake NetworkFee here - with a confirmationTime of '0'?

crypto/BRCryptoPayment.c
```diff
@@ -452,59 +477,144 @@ cryptoPaymentProtocolRequestGetPrimaryTargetAddress (BRCryptoPaymentProtocolRequ
...
 }
 
 static uint8_t*
 cryptoPaymentProtocolRequestGetMerchantData (BRCryptoPaymentProtocolRequest protoReq, size_t *merchantDataLen) {
+    uint8_t *merchantData = NULL;
     switch (protoReq->type) {
         case CRYPTO_PAYMENT_PROTOCOL_TYPE_BITPAY:
         case CRYPTO_PAYMENT_PROTOCOL_TYPE_BIP70: {
             BRPaymentProtocolRequest *request = protoReq->u.btc.request;
+            merchantData = request->details->merchantData;
```
EBGToo: The return type does not include 'const' but `details->merchantData` is not copied?  At the call site, it looks like 'const' would work for this function.

crypto/BRCryptoPayment.c
```diff
@@ -544,65 +654,72 @@ struct BRCryptoPaymentProtocolPaymentRecord {
...
-            uint64_t refundAmountInt = cryptoAmountGetIntegerRaw (baseAmount, &overflow);
-            // TODO(fix): Should this be an assert?
-            assert (CRYPTO_FALSE == overflow);
+            BRCryptoAmount baseAmount = cryptoAmountConvertToUnit (refundAmount, baseUnit);
+            if (NULL != baseAmount) {
+                BRCryptoBoolean overflow = CRYPTO_TRUE;
+                uint64_t refundAmountInt = cryptoAmountGetIntegerRaw (baseAmount, &overflow);
+
+                BRCryptoBoolean isAddressBTC = 0;
+                BRAddress reundAddressBtc = cryptoAddressAsBTC (refundAddress, &isAddressBTC);
```
EBGToo: Spelling: `reundAddr...` -> `refundAddr...`

crypto/BRCryptoPayment.c
michael-brd: Not fake, from the BitPay request. Fees have a time but I can't possible know the time as that isn't provided by the bitpay protocol.

crypto/BRCryptoWallet.c
michael-brd: Good catch. Will fix.

crypto/BRCryptoPayment.c
michael-brd: Good catch. Will fix.

crypto/BRCryptoPayment.c
EBGToo: Yes, surely a fee is required - but in this case it isn't related to the network's 'time to confirm'.

PR #181, opened 2019-08-29
CORE-509: Register BTC transactions upon successful submission

michael-brd:


bitcoin/BRSyncManager.c
```diff
@@ -857,7 +857,7 @@ BRClientSyncManagerSubmit(BRClientSyncManager manager,
                                 BRClientSyncManagerAsSyncManager (manager),
                                 (BRSyncManagerEvent) {
                                     SYNC_MANAGER_TXN_SUBMITTED,
-                                    { .submitted = {transaction->txHash, -1} },
+                                    { .submitted = {transaction, -1} },
```
EBGToo: The function declaration specifies `transaction` as `OwnershipKept` - does this event callback take the transaction?

This:
```
typedef void
(*BRSyncManagerEventCallback) (void * context,
                               BRSyncManager manager,
                               OwnershipKept BRSyncManagerEvent event);
```
implies that the contents of `event` are themselves kept?

PR #183, opened 2019-08-30
CORE-514: Report BTC/BCH P2P Sync Progress Every One Second

EBGToo:
@michael-brd Marked this as draft; is it is my notional implementation.


EBGToo:
CORE-519

PR #184, opened 2019-08-30
CORE-510: Process BDB 'next:href' links

EBGToo:
Previously, bdbMakeRequest() passed a 'more:Bool' flag to the completion handler; now it passes a 'more:URL?' to indicate if more data remains and the URL should be hit.


michael-brd:
Please create a JIRA issue to track the Java changes.

PR #185, opened 2019-08-30
CORE-515: Make System defaults consistent w/ BDB

EBGToo:
DEPENDS on CORE-512.  Will merge in `crypto` once CORE-512 is there.


michael-brd:
Please create a JIRA to track the required Java changes.

PR #187, opened 2019-09-03
CORE-518: Add 'instant sync' to Java Core Crypto

michael-brd:


EBGToo:
Looks like we need a Swift update for 'blocks'?  And Swift and Java updates for 'transfers'?

michael-brd:
> Looks like we need a Swift update for 'blocks'? And Swift and Java updates for 'transfers'?

Potentially, once required by a client I would say.

EBGToo:
CORE-529

PR #188, opened 2019-09-03
CORE-524: Update HTTP status code handling

michael-brd:
Draft for now, until api.blockset.com submission is restored.


Java/Crypto/src/main/java/com/breadwallet/crypto/blockchaindb/apis/bdb/BdbApiClient.java
```diff
@@ -281,26 +327,30 @@ public void handleError(QueryError error) {
         }
     }
 
-    private static class ArrayHandler<T> implements ResponseHandler {
+    private static class EmbeddedArrayResponseHandler<T> implements ResponseHandler<JSONObject> {
```
EBGToo: Nice!

EBGToo:
Okay.  I see the 'deserializer' exists for the one case where the BDB request succeeded but provided no data (like the 'create transaction' POST).

PR #190, opened 2019-09-04
CORE-522: BRFileService SQLite3 Implementation

EBGToo:
It works.

A number of things missing: a) not everything is 'released'; b) should remove 'ErrorHandler' in favor of 'state/error' returns on fileServiceLoad(), etc.; c) nothing on rollback, etc.  Might be able to add some BRFileServices functions to 'saveSet/saveArray' (matching Bitcoin blocks/peers callbacks).

Android would need to include a sqlite library.

Probably needs a name change away from 'File'...

Thoughts?


EBGToo:
Note: there is one database in <storage-path>/entities.db.  Each BRFileService instance opens a connection to that one database - thus 3 connections, one each for BTC, BCH, ETH.

There is one table 'Entity'.  The columns {Currency, Network} distinguish BTC, BCH, and ETH (just like the directory hierarchy did).  The additional columns {Type, Hash} distinguish the entity.

Hash is the primary key - no overlap in block, peers, transaction or log hashes (cryptographic hash, after all).  A true primary key would be the quad-tuple {currency, network, type, hash}

support/BRFileService.c
```diff
@@ -103,18 +168,38 @@ fileServiceEntityTypeAddHandler (BRFileServiceEntityType *entityType,
...
+}
+
 extern BRFileService
 fileServiceCreate (const char *basePath,
                    const char *currency,
                    const char *network,
                    BRFileServiceContext context,
                    BRFileServiceErrorHandler handler) {
+    if (NULL == currency || 0 == strlen(currency)) return NULL;
+    if (NULL == network  || 0 == strdup(network))  return NULL;
```
michael-brd: Leak on strdup (did you mean strlen?)

support/BRFileService.c
```diff
@@ -30,10 +30,75 @@
 #include <dirent.h>
 #include <sys/stat.h>
 #include <errno.h>
+#include <sqlite3.h>
```
michael-brd: Need to think about this; do we want iOS and Android using different versions of sqlite? Would we prefer that they use the same (included) version, at the cost of binary size?

support/BRFileService.c
```diff
@@ -127,21 +212,96 @@ fileServiceCreate (const char *basePath,
...
+    }
 
-    sprintf (dirPath, "%s/%s", basePath, currency);
-    if (-1 == directoryMake(dirPath)) return NULL;
+    // Create the SQLite 'Entity' Table
+    sqlite3_stmt *sdbCreateTableStmt;
+    status = sqlite3_prepare_v2 (fs->sdb, FILE_SERVICE_SDB_ENTITY_TABLE, -1, &sdbCreateTableStmt, NULL);
+    if (SQLITE_OK != status) {
+        // ...
+        printf ("foo\n");
```
michael-brd: Are these cases meant to be unhandled?

support/BRFileService.c
```diff
@@ -127,21 +212,96 @@ fileServiceCreate (const char *basePath,
...
+ //       return NULL;
+    }
+    sqlite3_finalize(sdbCreateTableStmt);
 
-    sprintf(dirPath, "%s/%s/%s", basePath, currency, network);
-    if (-1 == directoryMake(dirPath)) return NULL;
+    FileServiceSQL sql;
 
-    BRFileService fs = calloc (1, sizeof (struct BRFileServiceRecord));
+    // Create the SQLITE 'Insert into Entity' Statement
```
michael-brd: We are leaking, in all the return NULL cases.

support/BRFileService.c
```diff
@@ -152,9 +312,17 @@ fileServiceRelease (BRFileService fs) {
     size_t typesCount = array_count(fs->entityTypes);
     for (size_t index = 0; index < typesCount; index++)
         fileServiceEntityTypeRelease (&fs->entityTypes[index]);
+    array_free(fs->entityTypes);
+
+
+    free (fs->network);
+    free (fs->currency);
+    free (fs->sdbPath);
+    sqlite3_finalize(fs->sdbInsertStmt);
```
michael-brd: As you mentioned, not all statements are released.

support/BRFileService.c
```diff
@@ -265,104 +433,142 @@ fileServiceLoad (BRFileService fs,
...
-                (version != entityType->currentVersion ||
-                 headerVersion != currentHeaderFormatVersion))
-                fileServiceSave (fs, type, entity);
+    while (SQLITE_ROW == sqlite3_step(fs->sdbSelectAllStmt)) {
+        const char *hash = (const char *) sqlite3_column_text (fs->sdbSelectAllStmt, 0);
+        const char *data = (const char *) sqlite3_column_text (fs->sdbSelectAllStmt, 1);
+
+        // Hex-Decode hash
+        size_t  hashBytesCount = strlen(hash) / 2;
+        uint8_t hashBytes [hashBytesCount];
```
michael-brd: Are we okay using this pattern? I thought we were moving away from it in new code?

support/BRFileService.c
```diff
@@ -30,10 +30,75 @@
...
 
+#define FILE_SERVICE_SDB_FILENAME      "entities.db"
+
+#define FILE_SERVICE_SDB_ENTITY_TABLE     \
+"CREATE TABLE IF NOT EXISTS Entity(     \n\
+  Currency  CHAR(64)    NOT NULL,       \n\
+  Network   CHAR(64)    NOT NULL,       \n\
+  Type      CHAR(64)    NOT NULL,       \n\
+  Hash      TEXT        PRIMARY KEY,    \n\
+  Data      TEXT        NOT NULL);"
```
michael-brd: For Data, why not use BLOB? For the primary key, is using TEXT going to be a performance issue at some point (unlikely, but just asking)?

support/BRFileService.c
```diff
@@ -265,104 +433,142 @@ fileServiceLoad (BRFileService fs,
...
-            // Update restuls with the newly restored entity
-            BRSetAdd (results, entity);
-
-            // If the read version is not the current version, update
-            if (updateVersion &&
-                (version != entityType->currentVersion ||
-                 headerVersion != currentHeaderFormatVersion))
-                fileServiceSave (fs, type, entity);
+    while (SQLITE_ROW == sqlite3_step(fs->sdbSelectAllStmt)) {
+        const char *hash = (const char *) sqlite3_column_text (fs->sdbSelectAllStmt, 0);
```
michael-brd: Likely want to defensively treat all input from the FS (i.e. length checks, overflow checks, etc?).

michael-brd:
I started reviewing but realized its very much a PoC, so, apologies for the markup.

Concept looks good to me. Couple of thoughts:
* Threading (https://www.sqlite.org/threadsafe.html); what's the plan here?
* Versioning; most DB wrappers provide a pattern where the DB schema version is stored in the database (in a table) so that you can detect version upgrades (ex: https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/database/sqlite/SQLiteOpenHelper.java). Might we want to do the same?

support/BRFileService.c
```diff
@@ -30,10 +30,75 @@
...
 #include <errno.h>
+#include <sqlite3.h>
+typedef int sqlite3_status_code;
 
 #define FILE_SERVICE_INITIAL_TYPE_COUNT    (5)
 #define FILE_SERVICE_INITIAL_HANDLER_COUNT    (2)
 
+#define FILE_SERVICE_SDB_FILENAME      "entities.db"
+
+#define FILE_SERVICE_SDB_ENTITY_TABLE     \
```
michael-brd: Likely want to create an index on currency/network/type.

support/BRFileService.c
EBGToo: Yes.

support/BRFileService.c
EBGToo: Yes, pending error processing

support/BRFileService.c
EBGToo: yes.

support/BRFileService.c
EBGToo: I think IOS uses the built-in, 'blessed-by-Apple' version.  And Android should use the 'blessed-by-Google' version.  I certainly don't want to deal with upgrades and issue tracking on a version we choose.

support/BRFileService.c
EBGToo: Oops.

support/BRFileService.c
EBGToo: Yes, will remove that pattern.  And actually, a Hash is always UInt256 (32 bytes / 64 hex chars).  So, will flow that simplification through.

support/BRFileService.c
EBGToo: BLOB to avoid the hex encoding?  I do like seeing the hex-encoding in 'DB Query Tools' but admittedly have not seen how those tools (the one I use) displays BLOBs.

I know there is often concern about a non-integer PRIMARY KEY; but the Hash is the primary key; thus not much value in defining an unrelated integer value.  Is there a way to think of an integer as related?  Note: the actual primary key should be (Currency, Network, Type, Hash); can that be declared as such?

support/BRFileService.c
EBGToo: Sorry, missing the issue here.

michael-brd:
On 'threading', that page highlights that the default mode can be controlled at compile time. If we use the built in library, we do not have that control. That page highlights some avenues for setting the mode and/or detecting the default mode.

On 'version', yes, that would the case to handle. We could make the assumption that day never comes but it is worth thinking on/commenting on.

support/BRFileService.c
michael-brd: In terms of trusting the input values that we are parsing (potentially with just asserts).

support/BRFileService.c
michael-brd: You can have a primary key composed of multiple column values, IIRC.

michael-brd: You can run a query to display BLOBs as hex. I understand the desire to see as hex-encoding, but it does have a (small) performance/size implication.

support/BRFileService.c
michael-brd: SQLite is not exposed by the Android NDK. It is only available (assuming you are "following the rules") in the SDK (i.e. at the Java layer).

So, 'blessed-by-Google' version does not exist.

support/BRFileService.c
EBGToo: Sorry, still not sure of what 'input values'. `fs->sdbXYZStmt` or `hash`?  [I need to look at the whole code block?]

support/BRFileService.c
EBGToo: So what should we do?  Using SQLite3 was a test for performance of BRFileService; Adrian reports that there are no longer any slow ups. 

Do we have another non-sqlite option?  (Maintain a Set of {blocks, peers, transaction}; on any change, write the entire set out to one file (for each type) - like in plist or json format?)

If you include the Android SDK are the .o and .h unavailable (deep in the JAR/AAR-thingy)?

michael-brd: Maybe but I don't think we want to be doing that. We could like just dlopen the library if we wanted to get hacky.

I would think we would consider a) including SQLite or b) use callbacks to expose the platform's SQLite implmentation.

EBGToo:
This might need a 'migrateAsETH' interface - to move existing file-based {logs, transactions, ...} into an SQLite version.  The move would start by reading them; this PRs code removes the file-based reading code....

On the other hand, ETH doesn't use P2P and thus losing the logs+transactions would get reconstructed upon the first API-based sync.

michael-brd:
> This might need a 'migrateAsETH' interface - to move existing file-based {logs, transactions, ...} into an SQLite version. The move would start by reading them; this PRs code removes the file-based reading code....
> 
> On the other hand, ETH doesn't use P2P and thus losing the logs+transactions would get reconstructed upon the first API-based sync.

Good catch!

support/BRFileService.c
michael-brd: Hash and data. Are they 0 bytes, is dataBytes big enough to read from offset, as it grows, etc.

support/BRFileService.c
EBGToo: Should we bundle sqlite with Core?

michael-brd: Of the options, I'd say so. Assuming everyone is good with that, from a licensing/size/maintenance point of view.

As we've discussed, the file service could potentially be made to use callbacks so the platform sqlite libraries could be used. There are issues with this (possibly performance on Java, library versioning/behaviour, etc.).

We could also hedge a bit here and A) do the callback approach but B) also bundle in sqlite, with callbacks being provided for the bundled sqlite. Would give us flexibility, as well as (potentially, if we want to, needs more thought) use the platform sqlite instance where available (i.e. iOS) and the bundled one where not (easily i.e. Android).

Swift/BRCore.xcodeproj/project.pbxproj
```diff
@@ -535,6 +535,7 @@
 		3C590F5920950C750005597B /* BRBIP39Mnemonic.c */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.c; path = BRBIP39Mnemonic.c; sourceTree = "<group>"; };
 		3C5F5E73212C82310038B732 /* BREthereumMPT.h */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.h; path = BREthereumMPT.h; sourceTree = "<group>"; };
 		3C5F5E74212C82310038B732 /* BREthereumMPT.c */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.c; path = BREthereumMPT.c; sourceTree = "<group>"; };
+		3C6AE123231F013800016196 /* libsqlite3.tbd */ = {isa = PBXFileReference; lastKnownFileType = "sourcecode.text-based-dylib-definition"; name = libsqlite3.tbd; path = usr/lib/libsqlite3.tbd; sourceTree = SDKROOT; };
```
michael-brd: I think reference to libsqlite3's tbd file should be removed, ya?

Swift/BRCryptoDemo/CoreDemoAppDelegate.swift
```diff
@@ -77,18 +77,18 @@ class CoreDemoAppDelegate: UIResponder, UIApplicationDelegate, UISplitViewContro
...
-            }
-
-            try FileManager.default.createDirectory (atPath: storagePath,
-                                                     withIntermediateDirectories: true,
-                                                     attributes: nil)
-        }
-        catch let error as NSError {
-            print("Error: \(error.localizedDescription)")
-        }
+//        do {
```
michael-brd: Is this intended to be committed?

support/BRFileService.c
```diff
@@ -127,34 +228,158 @@ fileServiceCreate (const char *basePath,
...
+    // Create the file service itself
     BRFileService fs = calloc (1, sizeof (struct BRFileServiceRecord));
 
-    fs->pathToType = strdup(dirPath);
-    array_new (fs->entityTypes, FILE_SERVICE_INITIAL_TYPE_COUNT);
+    // A recursive lock - fileServiceLoad() will recursively call fileServiceSave()
+    {
+        pthread_mutexattr_t attr;
+        pthread_mutexattr_init(&attr);
+        pthread_mutexattr_settype(&attr, PTHREAD_MUTEX_RECURSIVE);
```
michael-brd: Recursive locks are a bit controversial. Would it be hard to remove its usage here? What is your opinion of them?

support/BRFileService.c
```diff
@@ -103,18 +177,45 @@ fileServiceEntityTypeAddHandler (BRFileServiceEntityType *entityType,
...
+    return NULL;
+}
+
 extern BRFileService
 fileServiceCreate (const char *basePath,
                    const char *currency,
                    const char *network,
                    BRFileServiceContext context,
                    BRFileServiceErrorHandler handler) {
+    if (NULL == currency || 0 == strlen(currency)) return NULL;
```
michael-brd: Do we want to throw a check on basePath in here too?

support/BRFileService.c
```diff
@@ -127,34 +228,158 @@ fileServiceCreate (const char *basePath,
...
+    // Save currency and network
+    fs->currency = strdup (currency);
+    fs->network  = strdup (network);
+
+    // Locate the SQLITE Database
+    size_t sdbPathLength = strlen (basePath) + 1 + strlen(currency) + 1 + strlen(network) + 1 + strlen (FILE_SERVICE_SDB_FILENAME) + 1;
+    fs->sdbPath = malloc (sdbPathLength);
+    sprintf (fs->sdbPath, "%s/%s-%s-%s", basePath, currency, network, FILE_SERVICE_SDB_FILENAME);
+
+    pthread_mutex_lock (&fs->lock);
```
michael-brd: Nothing wrong with it but curious, why do we lock in the "ctor"? Is a thread spawned off that could be accessing simultaneously?

support/BRFileService.c
```diff
@@ -127,34 +228,158 @@ fileServiceCreate (const char *basePath,
...
+// This is callable with a partially allocated BRFileService.  So, be
+// careful with fields that might not yet exist.
 extern void
 fileServiceRelease (BRFileService fs) {
-    size_t typesCount = array_count(fs->entityTypes);
-    for (size_t index = 0; index < typesCount; index++)
-        fileServiceEntityTypeRelease (&fs->entityTypes[index]);
+    pthread_mutex_lock (&fs->lock);
+    if (NULL != fs->entityTypes) {
+        size_t typesCount = array_count(fs->entityTypes);
```
michael-brd: There is the new `array_apply` and `array_free_all` that could be used here, if you were interested.

support/BRFileService.c
```diff
@@ -127,34 +228,158 @@ fileServiceCreate (const char *basePath,
...
+    if (NULL != fs->sdbDeleteAllTypeStmt) sqlite3_finalize (fs->sdbDeleteAllTypeStmt);
+    if (NULL != fs->sdbDeleteAllStmt)     sqlite3_finalize (fs->sdbDeleteAllStmt);
+
+    if (NULL != fs->sdb) sqlite3_close (fs->sdb);
+    pthread_mutex_unlock (&fs->lock);
+    pthread_mutex_destroy(&fs->lock);
 
-    free ((char *) fs->pathToType);
-    if (NULL != fs->entityTypes) array_free (fs->entityTypes);
     free (fs);
```
michael-brd: What do you think about, as a pattern, `memset`-ing to 0 in release functions? Allows us to more easily catch use after frees.

support/BRFileService.c
```diff
@@ -260,213 +508,259 @@ fileServiceLoad (BRFileService fs,
...
+        // Actually decode `data` into `dataBytes`
+        decodeHex (dataBytes, dataCount/2, data, dataCount);
+
+        size_t offset = 0;
+        BRFileServiceVersion version;
+        uint32_t  entityBytesCount;
+        uint8_t  *entityBytes;
+
+        BRFileServiceHeaderFormatVersion headerVersion = dataBytes[offset];
+        offset += 1;
```
michael-brd: I don't think we ensure that dataBytes is large enough before accessing it a offset throughout this loop? Maybe I missed, Github UI sucks.

michael-brd:
@EBGToo I pushed a commit to fix the Android build (and trim some of the fat I'd left in terms of options that we don't use). Looks good, ran it and a observed BTC transactions being loaded from disk. Nice work!

support/BRFileService.c
EBGToo: There is a recursive call to `fileServiceSave()`; you'd say avoid the RECURSIVE lock by splitting `fileServiceSave()` into a lock-free 'static' internal function?

support/BRFileService.c
EBGToo: Okay.

Swift/BRCryptoDemo/CoreDemoAppDelegate.swift
EBGToo: Yes.  But it changes in another, future commit.

support/BRFileService.c
michael-brd: Comes down to if we want to continue using recursive locks; if yes, no issue, if no, splitting would work.

Swift/BRCore.xcodeproj/project.pbxproj
EBGToo: I only see this in the .pbxprof file; not in any Xcode UI - I removed it in the UI previously.  I'll hand edit the file to remove it and see where that goes.

support/BRFileService.c
EBGToo: Fixed.

support/BRFileService.c
EBGToo: CORE-556

support/BRFileService.c
EBGToo: Hmmm, code has this `// Assert entityBytesCount remain in dataBytes`...  I'll fix.

support/BRFileService.c
EBGToo: Tried just now; compiler reported an error.

michael-brd: Ah, like a mismatch if you have an array of structs (instead of pointers) and the release function expects pointers. No big deal, was worth a shot.

support/BRFileService.c
EBGToo: Added `int needLock` parameter to a static version as `_fileServiceSave()`

support/BRFileService.c
```diff
@@ -127,34 +229,153 @@ fileServiceCreate (const char *basePath,
...
+
+    // Locate the SQLITE Database
+    size_t sdbPathLength = strlen (basePath) + 1 + strlen(currency) + 1 + strlen(network) + 1 + strlen (FILE_SERVICE_SDB_FILENAME) + 1;
+    fs->sdbPath = malloc (sdbPathLength);
+    sprintf (fs->sdbPath, "%s/%s-%s-%s", basePath, currency, network, FILE_SERVICE_SDB_FILENAME);
+
+    // Create/Open the SQLITE Database
+    sqlite3_status_code status = sqlite3_open(fs->sdbPath, &fs->sdb);
+    if (SQLITE_OK != status) {
+        pthread_mutex_unlock (&fs->lock);
```
michael-brd: I fear I created more work for you (sorry!). All the error paths still behave as if `fs->lock` is being held but I think you removed that now.
If its easier, just add that acquisition back in. Really was just a question on if it was needed.

support/BRFileService.c
```diff
@@ -260,213 +599,181 @@ fileServiceLoad (BRFileService fs,
...
+        if (updateVersion &&
+            (version != entityType->currentVersion ||
+             headerVersion != currentHeaderFormatVersion))
+            _fileServiceSave (fs, type, entity, 0);
     }
+    sqlite3_reset (fs->sdbSelectAllStmt);
+    pthread_mutex_unlock (&fs->lock);
 
-    if (0 != fclose (file)) {  fileServiceFailedUnix (fs, bytes, NULL, errno); return; }
+    if (dataBytes != dataBytesBuffer) free (dataBytes);
```
michael-brd: For the early return cases in the while loop, is this deallocation required there too? Same with the sqlite3_reset?

support/BRFileService.c
```diff
@@ -260,213 +599,181 @@ fileServiceLoad (BRFileService fs,
...
+    status = sqlite3_bind_text (fs->sdbDeleteStmt, 2, hash, -1, SQLITE_STATIC);
+    if (SQLITE_OK != status)
+        return fileServiceFailedSDB (fs, 1, status);
+
+    if (SQLITE_DONE != sqlite3_step (fs->sdbDeleteStmt))
+        return fileServiceFailedSDB (fs, 1, status);
 
-    char filename[strlen(fs->pathToType) + 1 + strlen(type) + 1 + 2*sizeof(UInt256) + 1];
-    sprintf (filename, "%s/%s/%s", fs->pathToType, type, u256hex(identifier));
+    sqlite3_reset (fs->sdbDeleteStmt);
```
michael-brd: Same comment re: is this sqlite3_reset needed in the early return cases?

support/BRFileService.c
```diff
@@ -260,213 +599,181 @@ fileServiceLoad (BRFileService fs,
...
-            remove (dirEntry->d_name); // If failed, then what?
+    status = sqlite3_bind_text (fs->sdbDeleteAllTypeStmt, 1, type, -1, SQLITE_STATIC);
+    if (SQLITE_OK != status)
+        return fileServiceFailedSDB (fs, 1, status);
 
-    closedir(dir);
+    if (SQLITE_DONE != sqlite3_step (fs->sdbDeleteAllTypeStmt))
+        return fileServiceFailedSDB (fs, 1, status);
+
+    sqlite3_reset (fs->sdbDeleteAllTypeStmt);
```
michael-brd: Same comment re: is this sqlite3_reset needed in the early return cases?

support/BRFileService.c
EBGToo: Fixed.

support/BRFileService.c
EBGToo: Early returns have the bytes freed if needed with:
```
            return fileServiceFailedEntity (fs, 1, (dataBytes == dataBytesBuffer ? NULL : dataBytes), NULL,
                                            type, "reader");
```

michael-brd: Nice!

support/BRFileService.c
EBGToo: I'll change these to be 'on entry' - ensure we start cleanly on each function entry.

support/BRFileService.c
EBGToo: As above.

PR #192, opened 2019-09-04
CORE-391: Do not re-query addresses when new ones are discovered

michael-brd:
Small change in Java to a) align the address count with Swift and b) order the params so that I can see the start/end numbers in logcat (were previously being cut off when they were at the end).

### DIFF
--- a/bitcoin/BRSyncManager.c
+++ b/bitcoin/BRSyncManager.c
@@ -91,6 +91,18 @@ struct BRSyncManagerStruct {
 
 /// MARK: - Sync Manager Decls & Defs
 
+struct BRClientSyncManagerScanStateRecord {
+    int requestId;
+    BRAddress lastExternalAddress;
+    BRAddress lastInternalAddress;
+    BRSetOf(BRAddress *) knownAddresses;
+    uint64_t begBlockNumber;
+    uint64_t endBlockNumber;
+    uint8_t isFullScan;
+};
+
+typedef struct BRClientSyncManagerScanStateRecord *BRClientSyncManagerScanState;
+
 struct BRClientSyncManagerStruct {
     // !!! must be first !!!
     struct BRSyncManagerStruct common;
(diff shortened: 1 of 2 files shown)


bitcoin/BRSyncManager.c
```diff
@@ -1232,62 +1221,178 @@ BRClientSyncManagerGenerateRid (BRClientSyncManager manager) {
...
+            *address = addressArray[index];
+            BRSetAdd (addresses, address);
+        }
+    }
+
+    free (addressArray);
+}
+
+static BRArrayOf(char *)
+_updateWalletAddressSet(BRSetOf(BRAddress *) addresses, BRWallet *wallet) {
```
EBGToo: This could return BRArrayOf(BRAddress) - the array_* macros will recognize BRAddress as a struct and allocate those chunks.  Doing so avoid the subsequence `malloc` in the loop.  I think.

I think a similar issue above in _filWalletAddressSet() - don't need to use `BRAddress *` and thus the `malloc()` (and subsequent `free()`

bitcoin/BRSyncManager.c
```diff
@@ -1093,77 +1114,49 @@ BRClientSyncManagerAnnounceGetTransactionsDone (BRClientSyncManager manager,
...
     }
 
-    if (addressStrings) {
-        free (addressStrings);
-    }
-
-    if (addressArray) {
-        free (addressArray);
+    if (NULL != addresses) {
+        array_free (addresses);
```
EBGToo: Does each 'char *' address need to be freed? BRSyncManagerGetTransactionsCallback is `OwnershipKept` for `addresses`.

bitcoin/BRSyncManager.c
michael-brd: I need an array of string pointers to pass to the BTC getTransactions call, not an array of BRAddress structs. Also, the BRSet holders pointers to items.

I also hoped to avoid having to do this but did not see a path (short of changing the getTransactions callback to expect BRAddress[], which I didn't think was a good idea).

Ultimately, this is C and the performance impact of allocations here isn't really an issue, I don't believe.

bitcoin/BRSyncManager.c
michael-brd: In this design, no, the individual address is owned by the scanState's knownAddress set. BUT it occurs to me that this callback occurs outside of the lock, so I will change it to avoid a race here.

michael-brd: `OwnershipKept` implies BRSyncManagerGetTransactionsCallback doesn't need to do anything with `addresses`, which is the current design. What am I missing from your comment in that respect?

EBGToo: I thought if `BRSyncManagerGetTransactionsCallback()` did nothing (by design) when BRSyncManager would need to do something with the array contents.  You confirmed that BRSyncManager owns them.

bitcoin/BRSyncManager.c
EBGToo: That is fine.  There are or can be BRSetOf 'BRAddress', 'BRAddress*' and 'char *'.  I was suggesting using 'BRAddress' for BRSet and then create BRArrayOf 'char *' from addr.s.  But, either is fine.

PR #193, opened 2019-09-05
CORE-531: Produce a SystemEvent When All Networks are Discovered

EBGToo:
This PR is 'double drafty'.

Probably the most important change is in BRCryptoWalletManager in the `create()` function (for ETH).  No longer is the ETH callback getTokens invoked; instead the network currencies (and the corresponding default unit) are used to a) create an ETH token and b) create an ETH wallet for that token.  The creation of the ETH wallet 'flows through the callbacks' to a) create an CWM wallet, and b) get announced via Swift/Java WalletManager and Wallet events.  [It may be that this change **is not** needed - it was originally made expecting (hoping) that the wallets would get announced immediately - that didn't happen]

Another change is to perform the system.configure action on the system.queue and to have that queue wait until all blockshains+currencies have produced their network.

### DIFF
--- a/Java/CoreCrypto/src/main/java/com/breadwallet/corecrypto/NetworkDiscovery.java
+++ b/Java/CoreCrypto/src/main/java/com/breadwallet/corecrypto/NetworkDiscovery.java
@@ -39,26 +39,22 @@ final class NetworkDiscovery {
 
     /* package */
     interface Callback {
-        void discovered(List<Network> networks,
-                        ImmutableMultimap<String, WalletManagerMode> updatedSupportedModes,
-                        ImmutableMap<String, WalletManagerMode> updatedDefaultModes);
+        void update(ImmutableMultimap<String, WalletManagerMode> updatedSupportedModes,
+                    ImmutableMap<String, WalletManagerMode> updatedDefaultModes);
+        void discovered(Network network);
+        void complete(List<com.breadwallet.crypto.Network> networks);
     }
 
     /* package */
     static void discoverNetworks(BlockchainDb query,
                                  boolean isMainnet,
+                                 List<com.breadwallet.crypto.blockchaindb.models.bdb.Currency> appCurrencies,
                                  ImmutableMultimap<String, WalletManagerMode> supportedModes,
                                  ImmutableMap<String, WalletManagerMode> defaultModes,
                                  Callback callback) {
-        List<Network> networks = new ArrayList<>();
+        List<com.breadwallet.crypto.Network> networks = new ArrayList<>();
 
-        ImmutableMultimap.Builder<String, WalletManagerMode> supportedModesBuilder = new ImmutableMultimap.Builder<>();
-        supportedModesBuilder.putAll(supportedModes);
-
-        ImmutableMap.Builder<String, WalletManagerMode> defaultModesBuilder = new ImmutableMap.Builder<>();
-        defaultModesBuilder.putAll(defaultModes);
-
-        CountUpAndDownLatch latch = new CountUpAndDownLatch(() -> callback.discovered(networks, supportedModesBuilder.build(), defaultModesBuilder.build()));
+        CountUpAndDownLatch latch = new CountUpAndDownLatch(() -> callback.complete(networks));
 
         getBlockChains(latch, query, isMainnet, blockchainModels -> {
             // Filter our defaults to be `self.onMainnet` and supported (non-nil blockHeight)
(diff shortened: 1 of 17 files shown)


Swift/BRCrypto/BRCryptoSystem.swift
```diff
@@ -496,107 +497,151 @@ public final class System {
...
-
-                                // Find the default unit
-                                let maximumDecimals = units.reduce (0) { max ($0, $1.decimals) }
-                                let defaultUnit = units.first { $0.decimals == maximumDecimals }!
-
-                                // Update associations
-                                associations[currency] = Network.Association (baseUnit: baseUnit,
-                                                                              defaultUnit: defaultUnit,
-                                                                              units: Set<Unit>(units))
+        // Query for blockchains on the system.queue - thus system.configure() returns instantly
```
michael-brd: To summarize the chunk of code change here, are there any business logic changes that Java need to mirror (aside from the new event) or is it just for the distpatch group/semaphore stuff?

EBGToo: Only threading stuff.

EBGToo:
I'm going to use this PR+JIRA to modify the `configure` interface with some type of `Set<Currencies>` argument.  Thus, continuing in 'draft' mode.

EBGToo:
And another change is to add `applicationCurrencies` to `System.configure` - thereby ensuring that if BDB is down, the we always configure with something (at least the built-in defaults and possibly augmented by those provided by the Application.

Swift/BRCrypto/BRCryptoSystem.swift
```diff
@@ -496,110 +499,194 @@ public final class System {
...
+
+                            // Find applicable defaults by `blockchainID`
+                            let defaults = System.defaultCurrencies
+                                .filter { $0.blockchainID == blockchainModel.id }
+
+                            // Find applicable application currencies by `blockchainID`
+                            let apps = applicationCurrencies
+                                .filter { $0.blockchainID == blockchainModel.id }
+
+                            // Merge in `defaults` with the result; but, on error, use apps
```
michael-brd: Does this give them the behaviour they want? Don't they want apps to always be merged in, otherwise they risk a wallet that the user expects?

EBGToo:
Requires Java changes.  @michael-brd you want to try the Java changes and see if they impact the Swift changes (either through the implementation or your thinking through of the implementation)?

michael-brd:
> Requires Java changes. @michael-brd you want to try the Java changes and see if they impact the Swift changes (either through the implementation or your thinking through of the implementation)?

@EBGToo Yes, will do now (parking networkIsReachable for now).

Swift/BRCrypto/BRCryptoSystem.swift
EBGToo: [Drat 'cancelled my review'; lost my comment here.]

I don't want to really on the currencies provided by the App - the BDB is the definitive source and the App should only be used as a stop-gap for a dead BDB.  That is part of the reason why the interface uses `[BlockChainDB.Model.Currency]` for the argument - it is a low-level thing substituting for a dead BDB.

Swift/BRCrypto/BRCryptoSystem.swift
EBGToo: Now includes 'supported' and `.discoveredNetworks` stuff.  And, if called multiple times will check for a pre-existing network before creating on.

crypto/BRCryptoWalletManager.c
```diff
@@ -461,6 +445,57 @@ cryptoWalletManagerRemWallet (BRCryptoWalletManager cwm,
...
+            switch (cwm->type) {
+                case BLOCK_CHAIN_TYPE_BTC:
+                    break;
+                case BLOCK_CHAIN_TYPE_ETH: {
+                    const char *address = cryptoCurrencyGetIssuer(c);
+                    if (NULL != address) {
+                        BREthereumGas      ethGasLimit = gasCreate(TOKEN_BRD_DEFAULT_GAS_LIMIT);
+                        BREthereumGasPrice ethGasPrice = gasPriceCreate(etherCreate(createUInt256(TOKEN_BRD_DEFAULT_GAS_PRICE_IN_WEI_UINT64)));
+
+                        tokenInstall (address,
```
michael-brd: Naive question; these gas limit / price values, what are the implications of using defaults? Are they updated later on? Does it matter if they are or are not?

crypto/BRCryptoWalletManager.c
```diff
@@ -461,6 +445,57 @@ cryptoWalletManagerRemWallet (BRCryptoWalletManager cwm,
...
 }
 
+static void
+cryptoWalletManagerInstallWalletsForCurrencies (BRCryptoWalletManager cwm) {
+    BRCryptoNetwork  network    = cryptoNetworkTake (cwm->network);
+    BRCryptoCurrency currency   = cryptoNetworkGetCurrency(network);
+    BRCryptoUnit     unitForFee = cryptoNetworkGetUnitAsBase (network, currency);
+
+    size_t currencyCount = cryptoNetworkGetCurrencyCount (network);
+    for (size_t index = 0; index < currencyCount; index++) {
```
michael-brd: Really like how this function is structured (w.r.t. gives/takes)

Swift/BRCrypto/BRCryptoSystem.swift
```diff
@@ -509,9 +509,11 @@ public final class System {
     /// `Network` there will be `SystemEvent` which can be used by the App to create a
     /// `WalletManager`
     ///
-    /// @Note: This should only be called one.
+    /// - Parameter applicationCurrencies: If the BlockChainDB does not return any currencies, then
+    ///     use `applicationCurrencies` merged into the deafults.  Appropriate currencies can be
+    ///     created from `System::asBlockChainDBModelCurrency` (see below)
     ///
-    public func configure () {
+    public func configure (withCurrencyModels applicationCurrencies: [BlockChainDB.Model.Currency]) {
```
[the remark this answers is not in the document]
EBGToo: I wanted to be clear-is in the interface that the `applicationCurrencies` are backups for what the BDB provides.  And, I didn't want to publicly expose the `Currencies` initializer.

That is one of the things that I thought would cause trouble in the Java...

I think it will be moot when 'tokens' use FileService.

crypto/BRCryptoWalletManager.c
EBGToo: Our 'estimate fee' interface should supersede these.

Java/CoreCrypto/src/main/java/com/breadwallet/corecrypto/NetworkDiscovery.java
```diff
@@ -142,15 +147,23 @@ static void discoverNetworks(BlockchainDb query,
                         return null;
                     }
 
-                    networks.add(Network.create(
+                    // define the network
+                    Network network = Network.create(
```
EBGToo: Is this `Network.create(...)` avoided if the system already has a network with the provided `id`?  The Swift code has a `guard` for that case and skipped out if the `id` is already known.

michael-brd: No, that guard occurs in System.java. I approached this a bit differently than you since Java is so verbose (didn't want it in System, want to break it out).

It does raise an interesting question, should the height from BDB be honoured in cases where the network is already known? Should the currencies? Fees?

EBGToo: I thought about the 'height' issue and concluded that the existing network will get its height from other defined interfaces, notably the periodic process.

Same thing with fees.  App is expected to call `System.updateNetworkFees`.

Currencies are another matter...

EBGToo: Created CORE-543

PR #194, opened 2019-09-06
CORE-535: Add granular BTC transaction lock

michael-brd:


EBGToo:
Since BWM has an EventHandler is it possible to make the BRPeerManger and BRWallet callbacks into BWM Events - with bwm{Signal,Handle}XYZ pairs - and then avoid the deadlock problem?  Doing so feels more general than a `transactions lock`.

I know some of the BR{Wallet,PeerManager} callbacks have some passed parameters (like blocks, peers) that have stack allocation which would necessitate a copy if used with bwm{Signal,Handle} pairs.  Copying might be worth it or as those callbacks don't touch transactions perhaps they need not be changed.

michael-brd:
> Since BWM has an EventHandler is it possible to make the BRPeerManger and BRWallet callbacks into BWM Events - with bwm{Signal,Handle}XYZ pairs - and then avoid the deadlock problem? Doing so feels more general than a `transactions lock`.
> 
> I know some of the BR{Wallet,PeerManager} callbacks have some passed parameters (like blocks, peers) that have stack allocation which would necessitate a copy if used with bwm{Signal,Handle} pairs. Copying might be worth it or as those callbacks don't touch transactions perhaps they need not be changed.

You've specifically expresses reservations about moving file I/O into a thread. That is why I pursued this design. Working with shifting requirements like is frustrating.

PR #195, opened 2019-09-06
CORE-498: Add sync depth support

michael-brd:


michael-brd:
@EBGToo I added the ETH implementation too but would appreciate a particularly critical eye on it.

bitcoin/BRSyncManager.c
```diff
@@ -1820,3 +1883,127 @@ _BRPeerSyncManagerTxPublished (void *info,
...
+
+    switch (depth) {
+        case SYNC_DEPTH_FROM_LAST_CONFIRMED_SEND: {
+            scanHeight = _getLastConfirmedSendTxHeight (wallet,
+                                                        networkBlockHeight,
+                                                        confirmationsUntilFinal);
+            break;
+        }
+        case SYNC_DEPTH_FROM_LAST_TRUSTED_BLOCK: {
+            const BRCheckPoint *checkpoint = BRChainParamsGetCheckpointBeforeBlockNumber (chainParams,
```
michael-brd: Does it make sense to use checkpoints when in API mode? I don't think there is any harm personally.

bitcoin/BRSyncManager.c
```diff
@@ -1820,3 +1883,127 @@ _BRPeerSyncManagerTxPublished (void *info,
...
+            array_add (newAddresses, address);
+        }
+    }
+
+    free (addressArray);
+    return newAddresses;
+}
+
+static uint32_t
+_getLastConfirmedSendTxHeight(BRWallet *wallet,
```
michael-brd: This isn't safe because the wallet can arbitrarily remove transactions while were are looping here. I am going to factor this out particular function into the BRWalletManager. Shouldn't affect the rest though.

michael-brd: This has been fixed as of commit id ea03c6c. That commit requires the changes in my other PR (https://github.com/breadwallet/breadwallet-core/pull/194) to be put through.

Java/Crypto/src/main/java/com/breadwallet/crypto/WalletManagerSyncDepth.java
```diff
@@ -0,0 +1,37 @@
...
+     * blockchain and mode as to how it determines trust.
+     */
+    FROM_LAST_TRUSTED_BLOCK,
+
+    /**
+     * Sync from the block height of the point in time when the account was created.
+     */
+    FROM_CREATION;
+
+    public int toSerialization() {
```
EBGToo: Thanks for this approach.

Swift/BRCrypto/BRCryptoWalletManager.swift
```diff
@@ -511,6 +515,60 @@ public enum WalletManagerMode: Equatable {
...
+///
+/// - fromCreation: Sync from the block height of the point in time when the account was created.
+///
+public enum WalletManagerSyncDepth: Equatable {
+    case fromLastConfirmedSend
+    case fromLastTrustedBlock
+    case fromCreation
+
+    /// Allow WalletMangerMode to be saved
+    public var serialization: UInt8 {
```
EBGToo: And thanks for this.

support/BRSyncMode.h
```diff
@@ -59,6 +59,24 @@ typedef enum {
...
+     * Sync from the block height of the last trusted block; this is dependent on the
+     * blockchain and mode as to how it determines trust.
+     */
+    SYNC_DEPTH_FROM_LAST_TRUSTED_BLOCK,
+
+    /**
+     * Sync from the block height of the point in time when the account was created.
+     */
+    SYNC_DEPTH_FROM_CREATION
+} BRSyncDepth;
```
EBGToo: Can you move this after the `BRSyncModeString()` function?

Swift/BRCrypto/BRCryptoWalletManager.swift
```diff
@@ -511,6 +515,60 @@ public enum WalletManagerMode: Equatable {
...
+/// The WalletManager's sync depth determines the range that a sync is performed on.
+///
+/// - fromLastConfirmedSend: Sync from the block height of the last confirmed send transaction.
+///
+/// - fromLastTrustedBlock: Sync from the block height of the last trusted block; this is
+///      dependent on the blockchain and mode as to how it determines trust.
+///
+/// - fromCreation: Sync from the block height of the point in time when the account was created.
+///
+public enum WalletManagerSyncDepth: Equatable {
```
EBGToo: So the IOS/Android logic is to 'increase the depth' based on User input (within a 24 hour period).  That is, start shallow, go deeper until the issue is resolved.  This interface does not have an implied ordering.  Perhaps something like:

```
var deeper: WalletManagerSyncDepth? { ... }    // stretching vocabulary
var shallower: WalletManagerSyncDept? { ... }
```

bitcoin/BRSyncManager.c
EBGToo: I'd do it as it parallels P2P mode as one less thing: to document, to explain to 'customer service' about recovery, to account for in test event/log sequences.

Swift/BRCrypto/BRCryptoWalletManager.swift
michael-brd: The ordering seems a bit goofy, as nothing is to say that last confirmed send will be before or after the last checkpoint. Regardless, added.

PR #199, opened 2019-09-10
CORE-527: Support `.api_only` for all blockchains.

EBGToo:
Requires Java changes.

### DIFF
--- a/Swift/BRCrypto/BRCryptoSystem.swift
+++ b/Swift/BRCrypto/BRCryptoSystem.swift
@@ -199,14 +199,21 @@ public final class System {
     ///
     /// Wallet Manager Modes
     ///
-
+    /// Every blockchain supports `.api_only`; blockchains with built-in P2P support (BTC, BCH,
+    /// and ETH) may support `.p2p_only`.  Intermediate modes (.api_with_p2p_submit,
+    /// .p2p_with_api_sync) are suppored on a case-by-case basis.
+    ///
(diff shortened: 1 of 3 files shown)


Swift/BRCrypto/BRCryptoSystem.swift
```diff
@@ -199,14 +199,21 @@ public final class System {
...
+    ///
+    /// It is possible that the `.api_only` mode does not work - for exmaple, the BDB is down.  In
+    /// that case it is an App issue to report and resolve the issue by: waiting out the outage;
+    /// selecting another mode if available.
+    ///
     var supportedModesMap: [String:[WalletManagerMode]] = [
-        "bitcoin-mainnet":      [.p2p_only],
-        "bitcoincash-mainnet":  [.p2p_only],
+        "bitcoin-mainnet":      [.api_only, .p2p_only],
+        "bitcoincash-mainnet":  [.api_only, .p2p_only],
```
michael-brd: Is BCH in  BDB?

EBGToo: So do we track BDB rollouts or do we let the WalletManager sync fail (API_ONLY selected, BDB produces no results or errors)?

michael-brd: I think releases lock down what they support. That way we limit potential unexpected behaviour.

EBGToo: CORE-542 created.

PR #200, opened 2019-09-11
CORE-538: Expose function to set if the network is reachable

michael-brd:
@EBGToo, I've updated the Java demo app to announce connectivity changes. I also exposed setNetworkReachable on both System and WalletManager.

I had thought to originally only have it on System but then the case comes up where a WalletManager comes in later on. We *could* have the System cache the value and apply it all created WalletManagers but then it gets a bit "stateful". Thought ultimately this was the most direct way to do it.

BRWalletManager defaults to TRUE for networkIsReachable, rather than expose it in the BRWalletManagerNew as a param (that function already has a ton of params). The comment highlights that this is in the (likely) case where a client doesn't bother keeping the flag up to date (ex: if this is running on a server that doesn't care/monitor network status).

### DIFF
--- a/Java/CoreCrypto/src/main/java/com/breadwallet/corecrypto/System.java
+++ b/Java/CoreCrypto/src/main/java/com/breadwallet/corecrypto/System.java
@@ -291,6 +291,8 @@ final class System implements com.breadwallet.crypto.System {
     private static final BRCryptoCWMListenerWalletEvent CWM_LISTENER_WALLET_CALLBACK = System::walletEventCallback;
     private static final BRCryptoCWMListenerTransferEvent CWM_LISTENER_TRANSFER_CALLBACK = System::transferEventCallback;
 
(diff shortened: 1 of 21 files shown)


EBGToo:
For the `WalletManager` function, the App is expected to call it?  If so, when?  I agree w/ not keeping the state in System.

michael-brd:
> For the `WalletManager` function, the App is expected to call it? If so, when? I agree w/ not keeping the state in System.

The demo app shows how (I think) it would be used; specifically, I use it on the `managerAdded` event. I get the current connectivity state and call `WalletManager.setNetworkReachable()` with it.

I use the `System.setNetworkReachable()` in response to changes announced by the system.

This approach *should* ensure that all WalletManagers have the correct flag value.

michael-brd:
As directed, changed System to track the network reachable state.

EBGToo:
Leaving it as `internal` is good; in the event a per-WalletManger reachability surfaces.  Thanks.

PR #206, opened 2019-09-11
CORE-532: In Core Ethereum, Properly Free Intermediate Sets on Error

EBGToo:
On an error, when initializing transactions, logs, blocks and nodes, be sure to call the appropriate 'release' functions as the Set is freed.

I tested this by adding the following compare with "logs" in fileServiceLoad():

            // Update results with the newly restored entity
            BRSetAdd (results, entity);

            if (0 == strcmp (type, "logs"))
                return 0;
and then confirming that:
a) the one log added got freed, and
b) all the prior transactions got freed.


EBGToo:
[Force-Pushed to get the commit comment to say 'CORE-545']

PR #212, opened 2019-09-13
CORE-537: Save ERC20 Tokens Persistently

EBGToo:
This had been on the back-burner.. but then CORE-554 happened.

I observed that in ewmCreate() there is this:
```
            FOR_SET (BREthereumLog, log, logs)
                ewmSignalLog (ewm, BCS_CALLBACK_LOG_ADDED, log);
```
signaling all logs.  And then in the handler there is this:
```
    BREthereumToken token = ewmLookupToken (ewm, logGetAddress(log));
    if (NULL == token) { logRelease(log); return;}
```
throwing away all the logs.

And then we wait for the next API sync to restore the logs after having installed the tokens.

This change moves the global `tokenSet` into `EWM` - a big win alone.  Tokens will exist immediately, from persistent storage, on `ewmCreate()`.  And then logs will be added to their appropriate wallet.  The balance is not updated based on the logs - that is in CORE-554.


ethereum/contract/BREthereumToken.c
```diff
@@ -117,6 +117,59 @@ tokenGetContract(BREthereumToken token) {
...
+extern void
+tokenUpdate (BREthereumToken token,
+             const char *symbol,
+             const char *name,
+             const char *description,
+             int decimals,
+             BREthereumGas defaultGasLimit,
+             BREthereumGasPrice defaultGasPrice) {
+
+    if (0 != strcasecmp (symbol     , token->symbol     )) { free (token->symbol     ); token->symbol      = strdup (symbol     ); }
```
michael-brd: General question; can these (name, description, symbol) be unicode? The standard `strxyz` don't support multibyte encodings, IIRC.

EBGToo: I've not thought about it.  Worthy of a JIRA to capture?

ethereum/ewm/BREthereumEWM.c
```diff
@@ -2955,3 +3051,48 @@ ewmTransferDelete (BREthereumEWM ewm,
...
+
+    if (NULL == token) {
+        token = tokenCreate (address,
+                             symbol,
+                             name,
+                             description,
+                             decimals,
+                             defaultGasLimit,
+                             defaultGasPrice);
+        BRSetAdd (ewm->tokens, token);
```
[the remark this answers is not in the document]
EBGToo: "Inconsistent Threading Model"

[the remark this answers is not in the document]
EBGToo: I added the `ewm->lock` over `ewm->tokens` *and* the update of the token itself.

ethereum/contract/BREthereumToken.c
michael-brd: Absolutely.

PR #216, opened 2019-09-14
CORE-560: Move 'secp256k1' to 'vendor/'; Remove BRCore (Xcode) module

EBGToo:
Still need to remove the 'secp256k1' git submodule (quasi-irreversible).


EBGToo:
I'm not sure how I got wedged on removing the submodule... Just need to move it to `vendor/`.  Will rework this.... duh.

PR #217, opened 2019-09-16
CORE-481: Add disconnect reason for the DISCONNECTED state

michael-brd:


michael-brd:
Still work to do:
* Add BRSyncStoppedReason?
* Add string message for posix reason (we don't want them calling strerror presumably)

@EBGToo, how is ETH able to advertise issues with connectivity/syncing?

michael-brd:
@EBGToo, this is ready for review. ETH doesn't return anything meaningful for both disconnects and sync stopped. Any thoughts on that front?

Swift/BRCrypto/BRCryptoWalletManager.swift
```diff
@@ -436,20 +436,39 @@ public final class WalletSweeper {
     }
 }
 
+public enum WalletManagerDisconnectReason: Equatable {
```
EBGToo: That is a good enum set.

Swift/BRCrypto/BRCryptoWalletManager.swift
```diff
@@ -590,6 +609,28 @@ public enum WalletManagerSyncDepth: Equatable {
     // Equatable: [Swift-generated]
 }
 
+public enum WalletManagerSyncStoppedReason: Equatable {
```
EBGToo: And so is this.

Swift/BRCrypto/BRCryptoWalletManager.swift
```diff
@@ -604,7 +645,7 @@ public enum WalletManagerEvent {
 
     case syncStarted
     case syncProgress (timestamp: Date?, percentComplete: Float)
-    case syncEnded (error: String?)
+    case syncEnded (reason: WalletManagerSyncStoppedReason)
```
EBGToo: Nice.

Swift/BRCrypto/common/BRSupport.swift
```diff
@@ -23,6 +23,10 @@ public func asUTF8String (_ chars: UnsafePointer<CChar>) -> String {
     return String (cString: chars, encoding: .utf8)!
 }
 
+public func asUTF8String (_ chars: UnsafePointer<UInt8>) -> String {
```
EBGToo: What is this?

support/BRSyncMode.h
```diff
@@ -33,6 +33,62 @@
...
+    DISCONNECT_REASON_UNKNOWN,
+    DISCONNECT_REASON_POSIX
+} BRDisconnectReasonType;
+
+typedef struct {
+    BRDisconnectReasonType type;
+    union {
+        struct {
+            int errnum;
+            char message[128 + 1];
```
EBGToo: Is `message` derived from `strerror(errno)`?  If so, there could be a function `const char * BRDisconnectReasonPosixGetMessage(...) { return strerror(...); }`.  Not a big deal either way.

EBGToo: Looks like it does come from `strerror()` and the existence of `message` requires some copying when filling out the `BR{Disconnect,SyncStopped)Reason` structures.

Swift/BRCrypto/common/BRSupport.swift
michael-brd: Bad leftover code from when I was trying to get the message out of the struct. Whoops! Good spot, will remove.

michael-brd:
I've made a complete hash of reworking this but it is now ready to be looked at again.

EBGToo:
> @EBGToo, this is ready for review. ETH doesn't return anything meaningful for both disconnects and sync stopped. Any thoughts on that front?

I see you used `BR*ReasonUnknown()` on EWM event callbacks into CWM.  I think that will have to be good enough for now.  `BREthereumState` is a simple enumerations and might need to follow our paradigm of: `enum ...StateType; struct { ...StateType type; union { ... } u; }`

The `BREthereumEWMEvent` structure does include an `errorDescription` field; but it is not actually used.  

I think we'll revisit ETH when it arrives on BDB.

EBGToo:
Good to go.

PR #224, opened 2019-09-17
CORE-572: malformed block data check

voisine:


EBGToo:
Ray ran the IOS app with the bad peer; we hit a break point in the new `if (off > bufLen) { ... }` and confirmed that `block->flagsLen` was huge and `off` was thus huge.  No subsequent crash.

PR #225, opened 2019-09-18
CORE-521: Add transfer submission errors and propagate them from C

michael-brd:


michael-brd:
@EBGToo, the `BRTransferSubmitError` (and associated `BRTransferSubmitErrorType`) types are defined in `support/BRSyncMode.h`.
I was hesitant (I'm not even sure why) to create a new file but they clearly have no business being there (same with `BRDisconnectReason`, actually).
How do we want to handle these types?

support/BRSyncMode.c
```diff
@@ -87,11 +92,76 @@ BRDisconnectReasonPosix(int errnum) {
...
+extern BRSyncStoppedReason
+BRSyncStoppedReasonPosix(int errnum) {
+    return (BRSyncStoppedReason) {
+        SYNC_STOPPED_REASON_POSIX,
+        { .posix = { errnum } }
+    };
+}
+
+extern const char *
+BRSyncStoppedReasonGetMessage(BRSyncStoppedReason *reason) {
```
EBGToo: Since this has a `strdup()` I think the return can't be `const`.  Also is worth noting that the return can be `NULL` - like in a documentation comment.  [Perhaps it is time for `\#define MIGHT_BE_NULLABLE` (name, obviously, TBD).]

Or don't `strdup()`.

EBGToo: Same with `BRDisconnectReasonGetMessage()`, `BRTransferSubmitErrorGetMessage()`, ...

crypto/BRCryptoWalletManagerClient.c
```diff
@@ -1217,16 +1220,11 @@ cwmTransactionEventAsETH (BREthereumClientContext context,
...
-                    { .included = {
-                        blockNumber,
-                        blockTransactionIndex,
-                        blockTimestamp,
-                        cryptoFeeBasisGetFee (feeBasisConfirmed)
-                    }}
-                };
-
+                // TODO(fix): BRCryptoTransferState leaks BRCryptoAmount
+                newState = cryptoTransferStateIncludedInit (blockNumber,
```
EBGToo: For the leak, this CORE-484

EBGToo:
> @EBGToo, the `BRTransferSubmitError` (and associated `BRTransferSubmitErrorType`) types are defined in `support/BRSyncMode.h`.
> 
> I was hesitant (I'm not even sure why) to create a new file but they clearly have no business being there (same with `BRDisconnectReason`, actually).
> 
> How do we want to handle these types?

I noticed that.  Understood the hesitancy to a new file and let it go.  But, we'll need to reconsider later - I'd like to unify many of the types used in BTC and ETH - which will naturally be in 'support' (as common to all blockchains).

Swift/BRCrypto/BRCryptoTransfer.swift
```diff
@@ -283,6 +283,23 @@ public class TransferHash: Hashable, CustomStringConvertible {
...
+    case posix(errno: Int32, message: String?)
+
+    internal init (core: BRTransferSubmitError) {
+        switch core.type {
+        case TRANSFER_SUBMIT_ERROR_UNKNOWN:
+            self = .unknown
+        case TRANSFER_SUBMIT_ERROR_POSIX:
+            var c = core
+            self = .posix(errno: core.u.posix.errnum,
+                          message: BRTransferSubmitErrorGetMessage (&c).map{ asUTF8String($0, true) } )
```
EBGToo: As `char *` returned, `true` is good.

Swift/BRCrypto/BRCryptoWalletManager.swift
```diff
@@ -450,7 +450,7 @@ public enum WalletManagerDisconnectReason: Equatable {
         case DISCONNECT_REASON_POSIX:
             var c = core
             self = .posix(errno: core.u.posix.errnum,
-                          message: BRDisconnectReasonPosixGetMessage(&c).map{ asUTF8String($0) })
+                          message: BRDisconnectReasonGetMessage(&c).map{ asUTF8String($0) })
```
EBGToo: This needs `asUTF8String` with `true`?

michael-brd: Yeah, fixed and pushed.

PR #228, opened 2019-09-18
CORE-561: In Core Crypto, Add WalletManager.registerWalletFor(currency).

EBGToo:
What is your feeling on this.  I think `requireWalletFor(...)` returning `Wallet?` is a bit awkward - unlike other interfaces.

It might be better to axe this function (as least publicly) and provide something at the `System` level.  Thinking...

Maybe add a `currencies: [Currency]` argument to `<system>.createWalletManager (..., currencies)`?  The `createWalletManager` call occurs in .networkAdded so all the currencies are known; they can be filtered by the App and then handled on CWM create.  <== This is better, no?

### DIFF
--- a/Swift/BRCrypto/BRCryptoSystem.swift
+++ b/Swift/BRCrypto/BRCryptoSystem.swift
@@ -318,7 +318,14 @@ public final class System {
     }
 
     ///
-    /// Create a wallet manager for `network` using `mode.
+    /// Create a wallet manager for `network` using `mode`, `addressScheme`, and `currencies.  A
+    /// wallet will be 'registered' for each of:
+    ///    a) the network's currency - this is the primaryWallet
+    ///    b) for each of `currences` that are in `network`.
+    /// These wallets are announced using WalletEvent.created and WalletManagerEvent.walledAdded.
+    ///
+    /// The created wallet manager is announed using WalletManagerEvent.created and
+    /// SystemEvent.managerAdded.
     ///
     /// - Parameters:
     ///   - network: the wallet manager's network
@@ -326,12 +333,18 @@ public final class System {
     ///   - addressScheme: the address scheme to use
     ///   - currencies: the currencies to 'register'.  A wallet will be created for each one.  It
     ///       is safe to pass currencies not in `network` as they will be filtered (but bad form
-    ///       to do so).
+    ///       to do so).  The 'primaryWallet', for the network's currency, is always created; if
+    ///       the primaryWallet's currency is in `currencies` then it is effectively ignored.
+    ///
+    /// - Note: There are two preconditions - `network` must support `mode` and `addressScheme`.
+    ///     Thus a fatal error arises if, for example, the network is BTC and the scheme is ETH.
     ///
     public func createWalletManager (network: Network,
                                      mode: WalletManagerMode,
                                      addressScheme: AddressScheme,
                                      currencies: Set<Currency>) {
+        precondition (supportsMode(network: network, mode))
+        precondition (supportsAddressScheme(network: network, addressScheme))
 
         let manager = WalletManager (system: self,
                                      callbackCoordinator: callbackCoordinator,
--- a/Swift/BRCrypto/BRCryptoSystem.swift
+++ b/Swift/BRCrypto/BRCryptoSystem.swift
@@ -322,11 +322,16 @@ public final class System {
     ///
     /// - Parameters:
     ///   - network: the wallet manager's network
-    ///   - mode: the mode to use
+    ///   - mode: the wallet manager mode to use
+    ///   - addressScheme: the address scheme to use
+    ///   - currencies: the currencies to 'register'.  A wallet will be created for each one.  It
+    ///       is safe to pass currencies not in `network` as they will be filtered (but bad form
+    ///       to do so).
     ///
     public func createWalletManager (network: Network,
                                      mode: WalletManagerMode,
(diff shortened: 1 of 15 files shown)


michael-brd:
I like your `<system>.createWalletManager (..., currencies)` idea a lot; let's do that!

EBGToo:
Implies Java changes.

Swift/BRCrypto/BRCryptoWalletManager.swift
```diff
@@ -232,6 +264,13 @@ public final class WalletManager: Equatable, CustomStringConvertible {
                    system: system,
                    callbackCoordinator: callbackCoordinator,
                    take: false)
+
+        // Register a wallet for each currency.
+        currencies
+            .forEach {
+                if network.hasCurrency ($0) {
+                    cryptoWalletManagerRegisterWallet (core, $0.core) }
```
michael-brd: Does this leak the returned wallet?

Swift/BRCryptoDemo/CoreDemoListener.swift
```diff
@@ -94,21 +99,29 @@ class CoreDemoListener: SystemListener {
...
                 let scheme = system.defaultAddressScheme(network: network)
 
+                let currencies = network.currencies
+                    .filter { (c) in registerCurrencyCodes.contains { c.code == $0 } }
+
                 let _ = system.createWalletManager (network: network,
                                                     mode: mode,
-                                                    addressScheme: scheme)
+                                                    addressScheme: scheme,
+                                                    currencies: currencies)
```
michael-brd: This feels tricky to me; we can't pass in the BTC/BCH/ETH "primary currencies", ya? This two level filtering ("networkCurrencyCodes"/"registerCurrencyCodes") is necessary? Can we just allow the "network currency codes" to work in the `cryptoWalletManagerRegisterWallet` function?

Maybe I am missing something.

michael-brd:
@EBGToo, I added the Java. This is a nice change. Couple small comments/observations.

Swift/BRCrypto/BRCryptoWalletManager.swift
EBGToo: Yes.  Will call `self.registerWalletFor(currency)` - doing the right thing *and* avoiding duplicate logic.

Swift/BRCryptoDemo/CoreDemoListener.swift
EBGToo: In the demo listener I use one set of currencyCodes for primary wallets to select the needed wallet manager and to specify the WalletManagerMode (the mode stuff is on some other branch somewhere, pending). The new, other set is used to register non-primary wallets.  

I think one set is okay.  Still need a '[String,WalletManagerMode]' map.

Swift/BRCrypto/BRCryptoSystem.swift
```diff
@@ -322,18 +322,24 @@ public final class System {
     ///
     /// - Parameters:
     ///   - network: the wallet manager's network
-    ///   - mode: the mode to use
+    ///   - mode: the wallet manager mode to use
+    ///   - addressScheme: the address scheme to use
+    ///   - currencies: the currencies to 'register'.  A wallet will be created for each one.  It
```
ewerx: Some of the expectations could be more explicit:
- it is assumed that network's native token wallet is created regardless, and this only affects wallet creation of non-native tokens
- including the native token in the set has no consequences (to avoid having to filter it out from `network.currencies`)

EBGToo: Updated to this:
```
    ///
    /// Create a wallet manager for `network` using `mode`, `addressScheme`, and `currencies.  A
    /// wallet will be 'registered' for each of:
    ///    a) the network's currency - this is the primaryWallet
    ///    b) for each of `currences` that are in `network`.
    /// These wallets are announced using WalletEvent.created and WalletManagerEvent.walledAdded.
    ///
    /// The created wallet manager is announed using WalletManagerEvent.created and
    /// SystemEvent.managerAdded.
    ///
    /// - Parameters:
    ///   - network: the wallet manager's network
    ///   - mode: the wallet manager mode to use
    ///   - addressScheme: the address scheme to use
    ///   - currencies: the currencies to 'register'.  A wallet will be created for each one.  It
    ///       is safe to pass currencies not in `network` as they will be filtered (but bad form
    ///       to do so).  The 'primaryWallet', for the network's currency, is always created; if
    ///       the primaryWallet's currency is in `currencies` then it is effectively ignored.
    ///
    /// - Note: There are two preconditions - `network` must support `mode` and `addressScheme`.
    ///     Thus a fatal error arises if, for example, the network is BTC and the scheme is ETH.
    ///
    public func createWalletManager (network: Network,
                                     mode: WalletManagerMode,
                                     addressScheme: AddressScheme,
                                     currencies: Set<Currency>) {
        precondition (supportsMode(network: network, mode))
        precondition (supportsAddressScheme(network: network, addressScheme))

        let manager = WalletManager (system: self,
                                     callbackCoordinator: callbackCoordinator,
                                     account: account,
                                     network: network,
                                     mode: mode,
                                     addressScheme: addressScheme,
                                     currencies: currencies,
                                     storagePath: path,
                                     listener: cryptoListener,
                                     client: cryptoClient)
        manager.setNetworkReachable(isNetworkReachable)
        self.add (manager: manager)
    }
```
Adding some preconditions.  Reasonable?

PR #229, opened 2019-09-19
CORE-526: In Core Crypto, Ensure EventHandler pthread_cond_{wait,signal} hold the same lock.

EBGToo:
See JIRA CORE-526 comment.


michael-brd:
I think you still have a race; with the gap between `eventHandlerTimeToQuit` and `eventQueueDequeue`.

I added comments to the JIRA. Really think we should take a look at settable events.

ethereum/event/BREvent.c
```diff
@@ -168,48 +142,45 @@ eventHandlerAlarmCallback (BREventHandler handler,
...
             }
 
+            case EVENT_STATUS_WAIT_ABORT:
+                timeToQuit = 1;
+                break;
+
             case EVENT_STATUS_NOT_STARTED:
             case EVENT_STATUS_UNKNOWN_TYPE:
             case EVENT_STATUS_NULL_EVENT:
+            case EVENT_STATUS_WAIT_ERROR:
```
michael-brd: Just based on the "impossible?" comment, do we want to assert on (some of?) these, to see if they occur?

ethereum/event/BREventQueue.c
```diff
@@ -82,6 +97,7 @@ eventQueueClear (BREventQueue queue) {
     queue->pending = NULL;
     queue->available = NULL;
 
+    pthread_cond_destroy(&queue->cond);
```
michael-brd: I don't think we want to destroy this here, but rather in `eventQueueDestroy`?

ethereum/event/BREventQueue.c
```diff
@@ -129,56 +146,105 @@ eventQueueEnqueue (BREventQueue queue,
...
-        this->next = queue->available;
-        queue->available = this;
-    }
+
+    BREventStatus status = EVENT_STATUS_SUCCESS;
+
+    pthread_mutex_lock (&queue->lock);
+    while (!queue->abort && !_eventQueueDequeue (queue, event))
+        if (0 != pthread_cond_wait (&queue->cond, &queue->lock))
+            return EVENT_STATUS_WAIT_ERROR;
```
michael-brd: This would represent a likely catastrophic failure so probably moot but early return here without releasing the `lock`.

ethereum/event/BREvent.c
```diff
@@ -168,48 +142,45 @@ eventHandlerAlarmCallback (BREventHandler handler,
...
 }
 
-
 typedef void* (*ThreadRoutine) (void*);
 
 static void *
 eventHandlerThread (BREventHandler handler) {
-
 #if defined (__ANDROID__)
     pthread_setname_np(handler->thread, handler->name);
```
michael-brd: I was thinking on the https://breadwallet.atlassian.net/browse/CORE-564 issue and it occurs to me that that same race *could* be a problem here (never actually observed).

If we change to `pthread_setname_np(pthread_self(), handler->name);`, I think we'd be safe here 😕? Threads suck.

ethereum/event/BREventQueue.c
EBGToo: Fixed.

ethereum/event/BREvent.c
EBGToo: Fixed.

ethereum/event/BREventQueue.c
EBGToo: Fixed, regarding a proper return path.

ethereum/event/BREvent.c
EBGToo: That is a change needed throughout.  I changed it... but it must be assigned in the function?

michael-brd: Its assigned in the thread calling `pthread_create` but this handler thread might hit that point before then (depending on cores/scheduling).

ethereum/event/BREvent.c
```diff
@@ -147,36 +147,36 @@ typedef void* (*ThreadRoutine) (void*);
 static void *
 eventHandlerThread (BREventHandler handler) {
 #if defined (__ANDROID__)
-    pthread_setname_np(handler->thread, handler->name);
+    pthread_setname_np (pthread_self(), handler->name);
 #else
-    pthread_setname_np(handler->name);
+    pthread_setname_np (pthread_self());
```
michael-brd: Doesn't `pthread_setname_np` take a `char *` as the first arg?

michael-brd:
Looks good to me. Fire in the hole, let's let this thing fly?

PR #233, opened 2019-09-19
CORE-525: Handle Pending Transactions

EBGToo:
Currently draft as we are waiting on BDB to announce submitted transactions as 'SUBMITTED' (currently we get nothing until 'included' even thought the transaction is properly forwarded to the BTC peers and in their mempools).


bitcoin/BRSyncManager.c
```diff
@@ -1162,25 +1164,23 @@ BRClientSyncManagerAnnounceGetTransactionsItem (BRClientSyncManager manager,
...
                 // If our transaction made it into the wallet, do not deallocate it
-                transaction = NULL;
+                needFree = 0;
             }
         }
     }
 
+    // Wallet already knows about this txn; so just update the block info.
+    if (/* isRegistered */ NULL != transaction && BRTransactionIsSigned (transaction)) {
+        BRWalletUpdateTransactions (manager->wallet, &transaction->txHash, 1, (uint32_t) blockHeight, (uint32_t) timestamp);
```
michael-brd: This should be in the above `needRegistration` if block, ya? (obviates the need to repeat the NULL and signed check, also respects the request ID being correct.

bitcoin/BRSyncManager.c
```diff
@@ -1279,7 +1279,9 @@ BRClientSyncManagerAnnounceGetTransactionsDone (BRClientSyncManager manager,
                                                       (const char **) addresses,
                                                       addressCount,
                                                       begBlockNumber,
-                                                      endBlockNumber,
+                                                      (BRClientSyncManagerScanStateIsFullScan (&manager->scanState)
```
michael-brd: Can this behaviour be folded into `BRClientSyncManagerScanStateGetEndBlockNumber`? The idea with all those `BRClientSyncManagerScanState` functions was to encapsulate, as much as possible, the scan behaviour/logic.

michael-brd: (alternatively, could set the `scanState->endBlockNumber` in `BRClientSyncManagerScanStateInit`, allowing `BRClientSyncManagerScanStateGetEndBlockNumber` to remain the same.

bitcoin/BRSyncManager.c
```diff
@@ -1353,7 +1355,9 @@ BRClientSyncManagerUpdateTransactions (BRClientSyncManager manager) {
...
```
michael-brd: Same comment re:moving to `BRClientSyncManagerScanStateInit`/`BRClientSyncManagerScanStateGetEndBlockNumber`.

support/BRBase.h
```diff
@@ -26,13 +26,34 @@
...
+
+// Inclusive bloch chain height bounds
+#define BLOCK_HEIGHT_MINIMUM       (0)
+#define BLOCK_HEIGHT_MAXIMUM       (UINT64_MAX - 5)  // Leave room for special values
+
+/**
+ * Check if `height` is between the MINIMUM and MAXIMUM values.
+ */
+#define BLOCK_HEIGHT_IS_IN_RANGE(height) \
+AS_CRYPTO_BOOLEAN (BLOCK_HEIGHT_MINIMUM <= (height) && (height) <= BLOCK_HEIGHT_MAXIMUM)
```
michael-brd: Might lead to dependency issues here (`AS_CRYPTO_BOOLEAN` comes from crypto/BRCryptoBase.h). Maybe we move `boolean` into support sooner rather than later?

bitcoin/BRSyncManager.c
```diff
@@ -1162,25 +1164,23 @@ BRClientSyncManagerAnnounceGetTransactionsItem (BRClientSyncManager manager,
...
-            transaction->blockHeight = (uint32_t) blockHeight;
-
+        if (NULL == BRWalletTransactionForHash (manager->wallet, transaction->txHash)) {
             // BRWalletRegisterTransaction doesn't reliably report if the txn was added to the wallet.
             BRWalletRegisterTransaction (manager->wallet, transaction);
-            if (BRWalletTransactionForHash (manager->wallet, transaction->txHash) == transaction) {
+            if (transaction == BRWalletTransactionForHash (manager->wallet, transaction->txHash)) {
                 // If our transaction made it into the wallet, do not deallocate it
-                transaction = NULL;
+                needFree = 0;
```
michael-brd: Nice.

Swift/BRCrypto/BRCryptoSystem.swift
```diff
@@ -1196,8 +1196,8 @@ extension System {
 
                 manager.query.getTransactions (blockchainId: manager.network.uids,
                                                addresses: addresses,
-                                               begBlockNumber: begBlockNumber,
-                                               endBlockNumber: endBlockNumber,
+                                               begBlockNumber: (begBlockNumber == BLOCK_HEIGHT_UNBOUND ? nil : begBlockNumber),
```
michael-brd: What does a nil begin block number map to, a scan from 0 presumably?

michael-brd:
@EBGToo, looks good. I updated the Java. I changed the Java wallet sweeper to use the `null` start/end height (rather than 0/`Network.height`). Does that seem like a reasonable change? Am I understanding correctly the BDB behaviour when the start number is omitted?

EBGToo:
> @EBGToo, looks good. I updated the Java. I changed the Java wallet sweeper to use the `null` start/end height (rather than 0/`Network.height`). Does that seem like a reasonable change? Am I understanding correctly the BDB behaviour when the start number is omitted?

I don't know the BDB behavior if `start_height` is omitted.  I made both 'start' and 'end' optional for uniformity reasons.

Swift/BRCrypto/BRCryptoSystem.swift
EBGToo: I don't know how BDB responds to a non-existent 'start_block', probably errors.

support/BRBase.h
EBGToo: I'll remove the `AS_CRYPTO_BOOLEAN` use.

Swift/BRCrypto/BRCryptoSystem.swift
michael-brd: > > @EBGToo, looks good. I updated the Java. I changed the Java wallet sweeper to use the `null` start/end height (rather than 0/`Network.height`). Does that seem like a reasonable change? Am I understanding correctly the BDB behaviour when the start number is omitted?
> 
> I don't know the BDB behavior if `start_height` is omitted. I made both 'start' and 'end' optional for uniformity reasons.

Okay, I walked back that change. Thanks!

michael-brd:
> > @EBGToo, looks good. I updated the Java. I changed the Java wallet sweeper to use the `null` start/end height (rather than 0/`Network.height`). Does that seem like a reasonable change? Am I understanding correctly the BDB behaviour when the start number is omitted?
> 
> I don't know the BDB behavior if `start_height` is omitted. I made both 'start' and 'end' optional for uniformity reasons.

Okay, I walked back that change. Thanks!

bitcoin/BRSyncManager.c
EBGToo: Don't we want to update the transaction even if we don't need to register it?

bitcoin/BRSyncManager.c
EBGToo: If so, please go ahead and change as you see fit.

michael-brd: Okay, give me 10.

bitcoin/BRSyncManager.c
EBGToo: I updated the logic with a clear check for if the update should occur.

EBGToo:
> Should code be lowercased here?

We got the case correct with the later commits, it seems.

PR #234, opened 2019-09-20
CORE-484: Fix memory leaks relating to BRCryptoTransfer

michael-brd:


michael-brd:
We had two leaks that are resolved in this. Since they occurred in overlapping areas, I dealt with the second one as well. Allowed me to actually confirm in Instruments that we no longer have any leaks for the use case of clicking around on the wallets/transfers (aside from the BRTxInputSetWitness and BRTxInputSetSignature faux-leaks in CORE-553).

The first leak is the known one regarding the fee for the INCLUDED state. The second one was from in `cwmTransactionEventAsETH`:

```
BRCryptoFeeBasis feeBasisConfirmed = cryptoFeeBasisCreateAsETH (cryptoTransferGetUnitForFee(transfer),
                                                                                feeBasisGetGasLimit(ethFeeBasis),
                                                                                feeBasisGetGasPrice(ethFeeBasis));
```

To deal with the transfer state being mutable, I added a lock to that struct. The only other field of that struct was the confirmed fee basis, so I wrapped that in the lock too.

crypto/BRCryptoWalletManagerClient.c
```diff
@@ -626,20 +628,21 @@ cwmTransactionEventAsBTC (BRWalletManagerClientContext context,
...
 
-                BRCryptoTransferState oldState = cryptoTransferGetState (transfer);
-                assert (CRYPTO_TRANSFER_STATE_INCLUDED != oldState.type);
-
-                // TODO(fix): BRCryptoTransferState leaks BRCryptoAmount; see CORE-484
                 BRCryptoTransferState newState = cryptoTransferStateIncludedInit (btcTransaction->blockHeight,
                                                                                   0,
                                                                                   btcTransaction->timestamp,
-                                                                                  cryptoFeeBasisGetFee (feeBasisConfirmed));
+                                                                                  fee);
```
EBGToo: So this is in part stylistic - but pushing towards consistency in how we pass arguments (regarding reference counts).

crypto/BRCryptoWalletManagerClient.c
```diff
@@ -667,30 +670,39 @@ cwmTransactionEventAsBTC (BRWalletManagerClientContext context,
...
+                cryptoTransferSetState (transfer, newState);
+
+                cwm->listener.transferEventCallback (cwm->listener.context,
+                                                     cryptoWalletManagerTake (cwm),
+                                                     cryptoWalletTake (wallet),
+                                                     cryptoTransferTake (transfer),
+                                                     (BRCryptoTransferEvent) {
+                                                         CRYPTO_TRANSFER_EVENT_CHANGED,
+                                                         { .state = { oldState, newState }}
+                                                     });
```
EBGToo: I see that `oldState` safely holds `fee` (from the change to `cryptoTransferGetState()`)

crypto/BRCryptoWalletManagerClient.c
```diff
@@ -1286,19 +1319,25 @@ cwmTransactionEventAsETH (BREthereumClientContext context,
...
     }
 
-    // TODO: crypto{Wallet,Transfer}Give()
+    if (NULL != transfer) {
+        cryptoTransferGive (transfer);
+    }
+
+    if (NULL != wallet) {
+        cryptoWalletGive (wallet);
+    }
```
EBGToo: Okay.  I see this whole function as being explicit, clearer in argument passing.   (In the switch cases one doesn't have to track "the last call shouldn't take" - just to avoid this final give.)

PR #236, opened 2019-09-20
CORE-583: In Core Crypto, the FileService is Account specific.

EBGToo:
Pending Java

### DIFF
--- a/Swift/BRCrypto/BRCryptoSystem.swift
+++ b/Swift/BRCrypto/BRCryptoSystem.swift
@@ -393,7 +393,7 @@ public final class System {
         self.listener  = listener
         self.account   = account
         self.onMainnet = onMainnet
-        self.path  = path
+        self.path  = path + (path.last == "/" ? "" : "/") + account.fileSystemIdentifier
         self.query = query
         self.listenerQueue = listenerQueue ?? DispatchQueue (label: "Crypto System Listener")
         self.callbackCoordinator = SystemCallbackCoordinator (queue: self.listenerQueue)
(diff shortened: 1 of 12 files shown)


Swift/BRCrypto/BRCryptoSystem.swift
```diff
@@ -393,7 +393,7 @@ public final class System {
         self.listener  = listener
         self.account   = account
         self.onMainnet = onMainnet
-        self.path  = path
+        self.path  = path + (path.last == "/" ? "" : "/") + account.fileSystemIdentifier
```
michael-brd: Do we think we'd ever support an environment where '/' isn't the (primary) separator (i.e. Windows)? Does Swift have something like Python's `os.path.sep` (that abstracts the actual value)?

Really, really small, likely moot question. Not meant to hold up review at all. Only change if you think it is worth it.

michael-brd: This actually led to an interesting quirk. In the Java demo app (I think the Swift too), when we re-create the System on a `reset`, we both read the path from the current system and use that for the new system.

Well, since we modify path, here, that means the path keeps growing ala:
* create initial system with path as "/test/"
* system sets path to "/test/<hash>"
* reset system, using the current system.path to pass to new system (i.e. "/test/<hash>")
* system sets path to "/test/<hash>/<hash>"
* reset again, using the current system.path to pass to new system (i.e. "/test/<hash>/<hash>")
* etc.

So, do we care? If so, do we want to leave path (as returned to the client) as is, and internally have theActualPath, which we don't return to the client?

I'll push the Java as is, then we can change (or not) both the Swift/Java depending on this discussion.

michael-brd:
> Pending Java

Added the Java implementation.

Swift/BRCrypto/BRCryptoSystem.swift
EBGToo: Pushed a 'fix'.  The App needs to pass in the same stuff on `reset`.  So I saved the stuff {account, storagePath, query, ...} in `CoreDemoAppDelegate`

PR #238, opened 2019-09-23
CORE-455: Migrate Cipher/Coder/Hasher/Signer implementations to C

michael-brd:
@EBGToo, before I go and implement all of the types, are you good with this approach?

For Swift, since it is an enum, is there a better way to handle the `cryptoHasherCreate`/`cryptoHasherGive`? Using an enum, can I store the `BRCryptoHasher` itself with the enum, rather than the `BRCryptoHasherType`, so that I don't have to create and release on each hash operation?


michael-brd:
@EBGToo, source-compatibility for Swift works!

At the C level, I changed "Encrypter" to "Cipher". Are you okay with that change?

Swift/BRCrypto/common/BRCryptoEncrypter.swift
```diff
@@ -17,161 +17,80 @@ public protocol Encrypter {
...
-            let keyCount = key.count
-            key.withUnsafeBytes { (keyBytes: UnsafeRawBufferPointer) -> Void in
-                let keyAddr  = keyBytes.baseAddress?.assumingMemoryBound(to: UInt8.self)
-
-                target = Data (source)
-                target.withUnsafeMutableBytes { (targetBytes: UnsafeMutableRawBufferPointer) -> Void in
-                    let targetAddr  = targetBytes.baseAddress?.assumingMemoryBound(to: UInt8.self)
-                    BRAESECBEncrypt(targetAddr, keyAddr, keyCount)
-                }
+    public static func chacha20_poly1305(key: Key, nonce12: Data, ad: Data) -> CoreEncrypter {
```
EBGToo: Nice.

PR #239, opened 2019-09-23
CORE-570: Don't perform `ewmUpdateWalletBalance()` on every periodic dispatch

EBGToo:
Only perform balance updates on the start of a sync (either a full sync or an incremental sync).


ethereum/ewm/BREthereumEWM.c
```diff
@@ -2719,10 +2732,10 @@ ewmPeriodicDispatcher (BREventHandler handler,
                 SUCCESS,
                 { .syncProgress = {
                     NO_SYNC_TIMESTAMP, // We do not have a timestamp
-                    AS_SYNC_PERCENT_COMPLETE(50) }}
+                    AS_SYNC_PERCENT_COMPLETE(66.67) }}
```
michael-brd: Few more PRs and we'll be at 100!

PR #249, opened 2019-09-25
CORE-604: Remove use of semaphores on executor service

michael-brd:


michael-brd:
@EBGToo, while looking at the BCH blocking issue this morning, I took a look at this (on the theory that the executor was being blocked).

We use a semaphore to execute a couple of operations (block/transaction retrieval). I've updated the Java to move away from that model. If you agree, I will create a ticket for a similar Swift change.

PR #250, opened 2019-09-25
CORE-456: Implement crypto primitive in Java Core Crypto

michael-brd:


Java/CoreCrypto/src/main/java/com/breadwallet/corecrypto/Encrypter.java
```diff
@@ -7,35 +7,57 @@
...
+
 /* package */
 final class Encrypter implements com.breadwallet.crypto.Encrypter {
 
-    // TODO(fix): Implement me!
-
     /* package */
     static Encrypter createForAesEcb(byte[] key) {
-        return null;
+        BRCryptoCipher cipher = BRCryptoCipher.createAesEcb(key).orNull();
```
EBGToo: I think we'd plan on renaming this class.  Worth doing now?

Java/CoreNative/src/main/java/com/breadwallet/corenative/crypto/BRCryptoHasher.java
```diff
@@ -0,0 +1,127 @@
...
+
+import java.nio.ByteBuffer;
+import java.util.Arrays;
+import java.util.List;
+
+import static com.google.common.base.Preconditions.checkState;
+
+public class BRCryptoHasher extends PointerType {
+
+    private static final int CRYPTO_HASHER_SHA1         = 0;
```
EBGToo: These need to agree with the C code?  If so, perhaps a comment; internal enum?

Java/CoreCrypto/src/main/java/com/breadwallet/corecrypto/Encrypter.java
michael-brd: Certainly. Do you want the Swift updated as well? Will cause a (easily resolvable) break in the iOS app.

michael-brd:
@EBGToo, I've updated to address the comments. Let me know if you are good with me merging in (small iOS breakage due to Encrypter->Cipher change).

EBGToo:
> @EBGToo, I've updated to address the comments. Let me know if you are good with me merging in (small iOS breakage due to Encrypter->Cipher change).

Yes, good.

PR #251, opened 2019-09-25
CORE-600: In Core Crypto Swift, on `System.init()` create the path for persistent storage.

EBGToo:
The System initializer become `init?()` for the case when the path cannot be created as a writable directory.


michael-brd:
Personal preference would be to not have system creation return an optional. As @ewerx commented, would prefer an assert/precondition/checkState.

Regardless, implemented the Java equivalent of the Swift's new behaviour.

EBGToo:
> Personal preference would be to not have system creation return an optional. As @ewerx commented, would prefer an assert/precondition/checkState.
> 
> Regardless, implemented the Java equivalent of the Swift's new behaviour.

I made is a precondition.

michael-brd:
> > Personal preference would be to not have system creation return an optional. As @ewerx commented, would prefer an assert/precondition/checkState.
> > Regardless, implemented the Java equivalent of the Swift's new behaviour.
> 
> I made it a precondition.

Updated.

PR #252, opened 2019-09-25
CORE-593: In Core Ripple, Rework gwmRecoverTransfer

carlcherry:
Reworked the function as per the JIRA item.  For Ripple
the function returns NULL and no transfers are show.

### DIFF
--- a/crypto/BRCryptoWalletManagerClient.c
+++ b/crypto/BRCryptoWalletManagerClient.c
@@ -1881,15 +1881,21 @@ cwmAnnounceGetTransactionsItemGEN (BRCryptoWalletManager cwm,
     cwm = cryptoWalletManagerTake (cwm);
 
     // Fundamentally, the `transfer` must allow for determining the `wallet`
-    BRGenericTransfer transfer = gwmRecoverTransfer (cwm->u.gen, transaction, transactionLength);
+    BRArrayOf(BRGenericTransfer) transfers = gwmRecoverTransfersFromRawTransaction (cwm->u.gen, transaction, transactionLength);
 
     // Announce to GWM.  Note: the equivalent BTC+ETH announce transaction is going to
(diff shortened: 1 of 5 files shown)


generic/BRGenericRipple.c
```diff
@@ -136,10 +136,10 @@ genericRippleWalletGetBalance (BRGenericWallet wallet) {
...
 }
 
-static BRGenericTransfer
-genericRippleWalletManagerRecoverTransfer (uint8_t *bytes,
-                                           size_t   bytesCount) {
-    return rippleTransactionCreateFromBytes (bytes, (int) bytesCount);
+static BRArrayOf(BRGenericTransfer)
+genericRippleWalletManagerRecoverTransfersFromRawTransaction (uint8_t *bytes,
+                                                            size_t   bytesCount) {
+    return NULL;
```
EBGToo: Reasonable.  Ripple will not event attempt to extract transfers from `bytes`.  Just return NULL here.

crypto/BRCryptoWalletManagerClient.c
```diff
@@ -1881,15 +1881,21 @@ cwmAnnounceGetTransactionsItemGEN (BRCryptoWalletManager cwm,
     cwm = cryptoWalletManagerTake (cwm);
 
     // Fundamentally, the `transfer` must allow for determining the `wallet`
-    BRGenericTransfer transfer = gwmRecoverTransfer (cwm->u.gen, transaction, transactionLength);
+    BRArrayOf(BRGenericTransfer) transfers = gwmRecoverTransfersFromRawTransaction (cwm->u.gen, transaction, transactionLength);
```
EBGToo: Although the parameters don't indicate it, `transaction` is owned by the caller and not to be freed here.  Check.

PR #264, opened 2019-09-27
CORE-586: In Core Crypto, Add System.destroy() and WalletManger.stop()

EBGToo:


EBGToo:
@michael-brd I added `fileSystemClose()` and invoke that during BWM, EWM, GWM stop functions.  In System, on `systemRemove` the removed system is now saved into an ever expanding array - to avoid the possibility of 'release'.  Callbacks are still filtered.  Made `System.wipe` public.  Removed `System.stopAll`; made `System.destroy` internal.

PR #269, opened 2019-09-30
CORE-627: In Core Crypto, Implement cryptoAmountAdd() for negative arguments

EBGToo:


Swift/BRCryptoTests/BRCryptoAmountTests.swift
```diff
@@ -116,8 +116,8 @@ class BRCryptoAmountTests: XCTestCase {
         XCTAssert (-1.5       == btc4.double (as: BTC_BTC))
         XCTAssert (-150000000 == btc4.double (as: BTC_SATOSHI))
 
-        XCTAssert ("-B1.50"          == btc4.string (as: BTC_BTC))
-        XCTAssert ("-SAT150,000,000" == btc4.string (as: BTC_SATOSHI))
+        XCTAssertEqual ("-B 1.50", btc4.string (as: BTC_BTC))
```
michael-brd: These tests all fail for me, with the addition of the space. Am I doing something wrong?

carlcherry: I didn't run the unit tests - but the changes to cryptoAmountAdd give me the correct result when calculating a balance.

EBGToo: The 'extra whitespace' appeared out-of-the-blue for me; apparently from the Apple NSNumberFormatter:
```
    private func formatterWith (unit: Unit) -> NumberFormatter {
        let formatter = NumberFormatter()
        formatter.locale = Locale.current
        formatter.numberStyle = .currency
        formatter.currencySymbol = unit.symbol;
        formatter.generatesDecimalNumbers = 0 != unit.decimals
        formatter.maximumFractionDigits = Int(unit.decimals)
        return formatter
    }
```
It is not actually a 'space' but some other whitespace.  It is failing for you in Xcode or in Android Studio?

michael-brd: Checking out this commit, running the tests in XCode, I get a failure. The added space seems to be the culprit.

michael-brd: I think I am using an iOS 12 simulator; perhaps this is tied to a change in iOS 13?

PR #275, opened 2019-10-02
CORE-506: Add System.updateNetworkFees

michael-brd:
@EBGToo, I noticed the Swift sets the networks fees at the Swift level only (I implemented the Java in the same way). Did we want to set those values at the C level?

### DIFF
--- a/Java/CoreCrypto/src/main/java/com/breadwallet/corecrypto/Network.java
+++ b/Java/CoreCrypto/src/main/java/com/breadwallet/corecrypto/Network.java
@@ -11,6 +11,7 @@ import android.support.annotation.Nullable;
 
 import com.breadwallet.corenative.crypto.CoreBRCryptoCurrency;
 import com.breadwallet.corenative.crypto.CoreBRCryptoNetwork;
+import com.breadwallet.corenative.crypto.CoreBRCryptoNetworkFee;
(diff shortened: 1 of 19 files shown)


EBGToo:
So this is the issue that the Network CTOR does:
```
       fees.forEach { cryptoNetworkAddNetworkFee (core, $0.core) }
```
thereby keeping `Network.fees` consistent with `BRCryptoNetwork`?  And the on update we get out-of-sync?

Instead `Network.fees` should be a 'computed property', mapping over (`get`) or installing (`set`) the Crypto C values?

michael-brd:
> So this is the issue that the Network CTOR does:
> 
> ```
>        fees.forEach { cryptoNetworkAddNetworkFee (core, $0.core) }
> ```
> 
> thereby keeping `Network.fees` consistent with `BRCryptoNetwork`? And the on update we get out-of-sync?
> 
> Instead `Network.fees` should be a 'computed property', mapping over (`get`) or installing (`set`) the Crypto C values?

Correct; C is out of sync with Java/Swift after an update.

EBGToo:
They should be in sync.

michael-brd:
> They should be in sync.

Agreed. Separate issue to add C functionality and update Swift/Java?

PR #276, opened 2019-10-02
CORE-617: In Core Crypto, cryptoWalletManagerCreate() can return NULL.

EBGToo:
Requires Java change.

### DIFF
--- a/crypto/BRCryptoWalletManager.c
+++ b/crypto/BRCryptoWalletManager.c
@@ -160,7 +160,7 @@ cryptoWalletManagerCreate (BRCryptoCWMListener listener,
                                              cwmPath,
                                              cryptoNetworkGetHeight(network),
                                              cryptoNetworkGetConfirmationsUntilFinal (network));
-            if (NULL == cwm->u.btc) { error = 1; goto cleanupOnError; }
+            if (NULL == cwm->u.btc) { error = 1; break ; }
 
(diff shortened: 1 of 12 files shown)


crypto/BRCryptoWalletManager.c
```diff
@@ -269,6 +275,12 @@ cryptoWalletManagerCreate (BRCryptoCWMListener listener,
         }
     }
 
+cleanupOnError:
```
michael-brd: Yikes! Goto statements? These can't be avoided? These would be the first in our source.

EBGToo: I know; maybe the first in my life:
```
static void
doCleanupOnWalletManagerError (BRCryptoWalletManager *cwmPtr) {
  cryptoWalletMangerGive (*cwmPtr);
  *cwmPtr = NULL;
}
```
and then
```
if (NULL = ....) { doCleanupOnWalletManagerError(&cwm); break; }
```
? 

Or goto?

michael-brd: Or just do (in the BTC case statement):
```
BRWalletManager bwm = BRWalletManagerNew();
if (NULL != bwm) {
  // normal business
} else {
  // error handling
}
```
Adds one level of indentation, but removes need for goto?

EBGToo: Updated.

PR #278, opened 2019-10-03
CORE-633: Fix fileServiceTypeBlockV1Reader to pass the block buffer l…

michael-brd:
…ength

### DIFF
--- a/Swift/CoreTests/CoreTests-Bridging-Header.h
+++ b/Swift/CoreTests/CoreTests-Bridging-Header.h
@@ -28,6 +28,11 @@ extern int BRRunTestWalletManagerSync (const char *paperKey,
                                        int isBTC,
                                        int isMainnet);
 
+extern int BRRunTestsBWM (const char *paperKey,
+                          const char *storagePath,
+                          int isBTC,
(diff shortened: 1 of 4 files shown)


michael-brd:
> Given that BRWalletManger.c contains:
> 
> ```
> static BRFileServiceTypeSpecification fileServiceSpecifications[]
> ```
> 
> as a `static` there isn't a way to add a test case to confirm the parsing? `#if defined (DEBUG) extern #else static #endif`?

That would would work. Since you authored the `fileService` "class", are you able to write the tests for the various operations?

EBGToo:
Let me give it a quick try; it would be in testBWM.c.  If it takes more than 1 hour, we'll add another JIRA for later.

EBGToo:
@michael-brd Added a test case covering blocks, transactions, peers.  Modified the writer/reader for BRPeer to ensure endian agnostic saving.

michael-brd:
> @michael-brd Added a test case covering blocks, transactions, peers. Modified the writer/reader for BRPeer to ensure endian agnostic saving.

Great! I'll merge on approval.

PR #283, opened 2019-10-03
CORE-634: In Core Ripple, Define BRGenericNetwork

carlcherry:
1. Add the BRGenericNetwork interfacei, functions, and handlers
2. Implement the single handler function "createAddress"
3. Add a BRGenericNetwork to BRCryptoNetwork
4. Modify cryptoNetworkCreateAddressFromString to use the generic
   network to create the address
5. Add a BRCryptoCurrency parameter to cryptoNetworkCreateAsGEN so
   that it can create the correct BRGenericNetwork.

### DIFF
--- a/crypto/BRCryptoNetwork.c
+++ b/crypto/BRCryptoNetwork.c
@@ -515,7 +515,7 @@ cryptoNetworkCreateAddressFromString (BRCryptoNetwork network,
 
         case BLOCK_CHAIN_TYPE_GEN:
         {
-            BRGenericAddress genericAddress = gwmNetworkCreateAddress(network->u.gen.net, string);
+            BRGenericAddress genericAddress = gwmNetworkAddressCreate(network->u.gen.net, string);
             return cryptoAddressCreateAsGEN (network->u.gen.net, genericAddress);
         }
     }
--- a/crypto/BRCryptoNetwork.c
+++ b/crypto/BRCryptoNetwork.c
@@ -139,6 +139,7 @@ struct BRCryptoNetworkRecord {
 
         struct {
             // TODO: TBD
+            BRGenericNetwork net;
             uint8_t mainnet;
         } gen;
     } u;
@@ -203,10 +204,13 @@ cryptoNetworkCreateAsETH (const char *uids,
 private_extern BRCryptoNetwork
 cryptoNetworkCreateAsGEN (const char *uids,
                           const char *name,
+                          BRCryptoCurrency currency,
                           uint8_t isMainnet) {
     BRCryptoNetwork network = cryptoNetworkCreate (uids, name);
     network->type = BLOCK_CHAIN_TYPE_GEN;
     network->u.gen.mainnet = isMainnet;
+    network->u.gen.net = gwmNetworkCreate(cryptoCurrencyGetCode(currency));
+    network->currency = cryptoCurrencyTake(currency);
     return network;
 }
 
@@ -241,6 +245,7 @@ cryptoNetworkRelease (BRCryptoNetwork network) {
             break;
     }
 
+    if (network->currency) cryptoCurrencyGive(network->currency);
     free (network->name);
     free (network->uids);
     if (NULL != network->currency) cryptoCurrencyGive (network->currency);
@@ -509,7 +514,10 @@ cryptoNetworkCreateAddressFromString (BRCryptoNetwork network,
             return cryptoAddressCreateFromStringAsETH (string);
 
         case BLOCK_CHAIN_TYPE_GEN:
-            return cryptoAddressCreateFromStringAsGEN (string);
+        {
+            BRGenericAddress genericAddress = gwmNetworkCreateAddress(network->u.gen.net, string);
+            return cryptoAddressCreateAsGEN(NULL, genericAddress);
+        }
     }
 }
 
--- a/crypto/BRCryptoNetwork.c
+++ b/crypto/BRCryptoNetwork.c
@@ -538,7 +538,7 @@ cryptoNetworkAsBTC (BRCryptoNetwork network) {
 private_extern void *
 cryptoNetworkAsGEN (BRCryptoNetwork network) {
     assert (BLOCK_CHAIN_TYPE_GEN == network->type);
-    return NULL;
+    return network->u.gen.net;
 }
 
 private_extern BRCryptoBlockChainType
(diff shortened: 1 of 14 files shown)


crypto/BRCryptoNetwork.c
```diff
@@ -203,10 +204,13 @@ cryptoNetworkCreateAsETH (const char *uids,
 private_extern BRCryptoNetwork
 cryptoNetworkCreateAsGEN (const char *uids,
                           const char *name,
+                          BRCryptoCurrency currency,
                           uint8_t isMainnet) {
     BRCryptoNetwork network = cryptoNetworkCreate (uids, name);
     network->type = BLOCK_CHAIN_TYPE_GEN;
     network->u.gen.mainnet = isMainnet;
+    network->u.gen.net = gwmNetworkCreate(cryptoCurrencyGetCode(currency));
+    network->currency = cryptoCurrencyTake(currency);
```
carlcherry: Not sure I need to keep the currency since it was only needed to create the BRGenericNetwork

crypto/BRCryptoNetwork.c
```diff
@@ -509,7 +514,10 @@ cryptoNetworkCreateAddressFromString (BRCryptoNetwork network,
             return cryptoAddressCreateFromStringAsETH (string);
 
         case BLOCK_CHAIN_TYPE_GEN:
-            return cryptoAddressCreateFromStringAsGEN (string);
+        {
+            BRGenericAddress genericAddress = gwmNetworkCreateAddress(network->u.gen.net, string);
+            return cryptoAddressCreateAsGEN(NULL, genericAddress);
```
carlcherry: This is the one issue remaining.  We have no way to get the generic wallet manager from here.  So either we find a way or we ensure that BRCryptoAddress can operate without one.

crypto/BRCryptoNetwork.c
EBGToo: I don't think you need it; doesn't hurt to hold it (as long as the 'cryptoNetworkRelease()` call `cryptoCurrencyGive(...)`

generic/BRGenericHandlers.h
```diff
@@ -33,6 +33,10 @@ extern "C" {
         BRGenericAccount base;
     } *BRGenericAccountWithType;;
 
+    typedef struct {
+        BRGenericHandlers handlers;
+    } BRGenericNetworkRecord;
+
```
EBGToo: This is okay for now but is something that needs consideration.

Notice that `BRGenericAddress`, `BRGenericWallet`, `BRGenericTransfer`, .. do not have corresponding structures  at the 'generic' level.  Instead address, wallet, transfer, ... are simply `void*` because they come from the 'ripple', 'stellar', ... level.  

What is the reason that `BRGenericAccountWithType` and `BRGenericNetworkRecord` need to exist?  It is because in these cases the `handlers` need to be accessed.  Normally the access to `handlers` is through BRGenericWalletManager.  But 'account' and 'network' can't know about a 'wallet manager' because 'account' and 'network' pre-exist 'wallet manager' and, in fact, are need to create a 'wallet manager'.

There is also a future issue of thread safety (lock/unlock).  The 'wallet manager' holds a lock that protects its contents including access to the 'wallet managers' wallet(s), transfers, etc.  Do 'account' and 'network' need a lock as well?

I'm actively wondering:

1. Perhaps these `typedef void*` declarations need to be replaced with a more uniform design like:
```
typedef struct ... {
  BRGenericHandlers handler;
  void *underlyingRippleStellar...Reference;
} BRGeneric<Thing>;
```
which, by holding `handlers`, allows the context for the underlying reference to be understood.
2. Or the CWM level, for things like 'account' and 'network' need to hold not only the 'void*' thing but also a 'handers' thing so that the context isn't lost.
3. Accept that 'account', 'network' and 'wallet manager' won't really be 'typedef void*', they will have some structure at the 'GEN' level, unlike wallet, transfer, etc.

@carlcherry, @michael-brd Am I making sense?  Thoughts?

generic/BRGenericHandlers.h
```diff
@@ -140,6 +148,10 @@ extern "C" {
             BRGenericFeeBasisIsEqual feeBasisIsEqual;
             BRGenericFeeBasisFree free;
         } feebasis;
+
+        struct {
+            BRGenericNetworkCreateAddress createAddress;
+        } network;
```
EBGToo: Nice, Ripple, Stellar, etc only need to create addresses; not networks to be created/released.

crypto/BRCryptoNetwork.c
EBGToo: See my comment below on `BRGenericNetworkRecord`.

generic/BRGenericHandlers.h
EBGToo: Although I think there needs to be `addressRelease` and ('createAddress' -> 'addressCreate' or 'networkCreateAddress')

generic/BRGenericHandlers.h
EBGToo: And more, the case of `Network` is an interesting one.  We've come to the conclusion, I think, that networks are built in, in some form.  Core software comes configured for {BTC Mainnet, BTC Testnet, BCH Mainnet, BCH Testnet, ETH Mainnet, ETH Ropsten, XRP Mainnet} - because we have source and configurations (e.g. ChainParams) built in.  Perhaps BRCryptoNetwork instances need to be 'statically defined' for the builtin networks/blockchains.

And 'Account' is also interesting.  In order to serialize an 'account' (tso the IOS/Android App can use it on restart) we *must* know about the built-in blockchains - because we *must* generate a public address from the BIP39 seed, while we have the seed and before use/serialization.

generic/BRGenericHandlers.h
carlcherry: Fixed in latest checkin

generic/BRGenericHandlers.h
michael-brd: I wonder if the statically defined bits should be split off into a BRCryptoBlockchain type (or what have you, which a BRCryptoNetwork would have). The BRCryptoBlockchain "class" would be immutable ("source and configurations"), while the network would have the mutable properties. The built-ins would then be of the BRCryptoBlockchain type, rather than BRCryptoNetwork.

That would provide a clean separation of responsibilities, rather than having built-in networks that we are not meant to mutate (i.e. cryptoNetworkSetHeight/cryptoNetworkSetNetworkFees).

Things like Account could then make due with the built-in blockchains.

EBGToo: Copied above to CORE-541.

PR #287, opened 2019-10-04
CORE-630: In Core Ripple, implement generic functions to support "sub…

carlcherry:
…mit"

Added generic functions to create a new transfer (and free)
Added generic functions to serialize and get bytes for a transfer
Implemented existing functions that are in the GEN submit path

### DIFF
--- a/Swift/BRCrypto/BRCryptoSystem.swift
+++ b/Swift/BRCrypto/BRCryptoSystem.swift
@@ -1753,6 +1753,8 @@ extension System {
                                                             let timestamp = transaction.timestamp.map { $0.asUnixTimestamp } ?? 0
                                                             let height    = transaction.blockHeight ?? 0
                                                             transaction.transfers.forEach { (transfer: BlockChainDB.Model.Transfer) in
+                                                                // TODO - see the note in JIRA item CORE-648 about what we do with
+                                                                // the following check (remove or assert)
                                                                 if (accountAddress == transfer.source ||
                                                                     accountAddress == transfer.target) {
                                                                     cwmAnnounceGetTransferItemGEN(cwm, sid, transaction.hash,
@@ -1768,6 +1770,8 @@ extension System {
                 }},
 
             funcSubmitTransaction: { (context, cwm, sid, transactionBytes, transactionBytesLength, hashAsHex) in
+                // TODO - figure out why this precondition failes - or just remove it since systemExtract
+                // does the exact same precondition: JIRA CORE-649
                 //precondition (nil != context  && nil != cwm)
 
                 guard let (system, manager) = System.systemExtract (context, cwm)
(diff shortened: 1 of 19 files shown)


Swift/BRCrypto/BRCryptoSystem.swift
```diff
@@ -1753,19 +1753,22 @@ extension System {
                                                             let timestamp = transaction.timestamp.map { $0.asUnixTimestamp } ?? 0
                                                             let height    = transaction.blockHeight ?? 0
                                                             transaction.transfers.forEach { (transfer: BlockChainDB.Model.Transfer) in
-                                                                cwmAnnounceGetTransferItemGEN(cwm, sid, transaction.hash,
-                                                                                              transfer.source, transfer.target,
-                                                                                              transfer.amountValue,
-                                                                                              transfer.amountCurrency,
-                                                                                              timestamp, height)
+                                                                if (accountAddress == transfer.source ||
```
carlcherry: We will assert later if we add a transfer that does not belong to us.  There seems to be a bug in BDB that gives us fee transfers for other users (it was fixed once but crept back in)

crypto/BRCryptoAddress.c
```diff
@@ -137,12 +137,6 @@ cryptoAddressCreateFromStringAsETH (const char *ethAddress) {
     return address;
 }
 
-extern BRCryptoAddress
```
carlcherry: No longer needed

crypto/BRCryptoWallet.c
```diff
@@ -681,20 +681,18 @@ cryptoWalletEstimateFeeBasis (BRCryptoWallet  wallet,
         }
 
         case BLOCK_CHAIN_TYPE_GEN: {
-            //            BRGenericWalletManager gwm = wallet->u.gen.gwm;
-            //            BRGenericWallet wid = wallet->u.gen.wid;
-            //
-            //            UInt256 genValue = cryptoAmountGetValue(amount);
-
-            // TODO: Generic EstimateFee
-            assert (0);
```
carlcherry: Get rid of assert for now so that the submit flow can work.

ripple/BRRippleFeeBasis.c
```diff
@@ -15,14 +15,12 @@
 
 extern BRRippleUnitDrops rippleFeeBasisGetPricePerCostFactor(BRRippleFeeBasis *feeBasis)
 {
-    assert(feeBasis);
```
carlcherry: for now - don't assert - just return a default

ripple/BRRippleTransfer.c
```diff
@@ -31,20 +32,45 @@ rippleTransferCreate(BRRippleAddress from, BRRippleAddress to,
...
+extern BRRippleTransfer /* caller must free - rippleTransferFree */
+rippleTransferCreateNew(BRRippleAddress from, BRRippleAddress to,
+                     BRRippleUnitDrops amount)
+{
+    BRRippleTransfer transfer = calloc (1, sizeof (struct BRRippleTransferRecord));
+    transfer->sourceAddress = from;
+    transfer->targetAddress = to;
+    transfer->amount = amount;
+    BRRippleFeeBasis feeBasis; // NOTE - hard code for DEMO purposes
+    feeBasis.pricePerCostFactor = 10;
```
carlcherry: Hard-coded for now.

crypto/BRCryptoWallet.c
carlcherry: One other thing.  I have another local change to Swift/BRCryptoDemo/TransferCreateSendController.swift that is not part of this change but is required for the demo: a default BRCryptoFeeBasis object since we don't have a callback to create one yet.  Should I check that in?

PR #292, opened 2019-10-07
CORE-646: In Core Crypto, Add 'common' checks to BRCryptoBaseTests for Swift

EBGToo:
Handle some BTC/ETH and API/P2P differences (but added a JIRA CORE-647 to reconcile).


EBGToo:
@michael-brd I pushed another commit to this; allowing the Demo App (Swift) and tests to run even if the Developer doesn't have a .brdCoreTestsConfig.json file - a default one is installed with testate and compromised mainnet paper keys.  Also removes the requirement to have an argument, of the paper key identifier, when invoking the Demo App.

Demo App really should query for the paper key on start - but that is too much to change at this point.

PR #295, opened 2019-10-08
CORE-629: Fix memory corruption issues

michael-brd:
I have some tests in `testBwm.c` that abuse the BRPeerManager (via the BRWalletManager). They call BRPeerManagerConnect and BRPeerManagerDisconnect, as well as BRPeerManagerDisconnect+BRPeerManagerFree. They also randomly set if the network is reachable or not.

Using these tests, I can reliably trigger user-after-frees on an ASAN build.

To get the discussion started on how to fix, I've got this PoC branch that "fixes" the issues. Fixes in quotations because this is probably not how we want to fix it.

Some of the issues/changes:
* `BRPeerManagerConnect` used to explicitly call `_peerDisconnected` after a race-y check against the BRPeer's thread; removed this
* Added asserts to capture pthread_create failures; these could/should be changed to the function returning success/failure and handling
* We close the BRPeer's socket in `BRPeerDisconnect`, but the BRPeer thread closes it too. We can't do that, as the file descriptor could possibly be reused. Since the BRPeer thread owns the socket (and has a close), removed the `close` in `BRPeerDisconnect`. Instead, just a `shutdown` is performed to signal our intent to close
* The BRPeerManager code sort of assumes that a BRPeer has a thread (pre-emptively increments peerThreadCount, only calls BRPeerFree in `_peerDisconnected`, etc.) but if the network is unreachable, no BRPeer thread is created. I removed the `waitingForNetwork` BRPeer field and rolled it into the status. On `BRPeerDisconnect`, if we never launched the thread (i.e. in the `BRPeerStatusWaiting` state), launch a thread that does the same actions a "real" BRPeer thread would do (i.e. call disconnect, do the thread cleanup).
* Added asserts to capture future issue with peer thread count "underflow"-ing

### DIFF
--- a/bitcoin/BRPeer.c
+++ b/bitcoin/BRPeer.c
@@ -99,7 +99,6 @@ typedef struct {
     uint32_t magicNumber;
     char host[INET6_ADDRSTRLEN];
     BRPeerStatus status;
-    int waitingForNetwork;
     volatile int needsFilterUpdate;
     uint64_t nonce, feePerKb;
     char *useragent;
@@ -982,7 +981,7 @@ static double _peerGetMempoolTime (BRPeerContext *ctx) {
 }
 
 
-static void *_peerThreadRoutine(void *arg)
+static void *_peerThreadConnectRoutine(void *arg)
 {
     BRPeer *peer = arg;
     BRPeerContext *ctx = arg;
@@ -1111,6 +1110,22 @@ static void *_peerThreadRoutine(void *arg)
     return NULL; // detached threads don't need to return a value
 }
 
+static void *_peerThreadDisconnectRoutine(void *arg) {
+    BRPeer *peer = arg;
+    BRPeerContext *ctx = arg;
+    
+    pthread_cleanup_push(ctx->threadCleanup, ctx->info);
+
+    peer_log(peer, "waiting-disconnected");
+
+    assert (0 == array_count(ctx->pongCallback));
+    assert (NULL == ctx->mempoolCallback);
+
+    if (ctx->disconnected) ctx->disconnected(ctx->info, 0);
+    pthread_cleanup_pop(1);
+    return NULL; // detached threads don't need to return a value
+}
+
 static void _dummyThreadCleanup(void *info)
 {
 }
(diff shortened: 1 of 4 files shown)


bitcoin/BRPeer.c
```diff
@@ -1111,6 +1110,22 @@ static void *_peerThreadRoutine(void *arg)
...
+static void *_peerThreadDisconnectRoutine(void *arg) {
+    BRPeer *peer = arg;
+    BRPeerContext *ctx = arg;
+    
+    pthread_cleanup_push(ctx->threadCleanup, ctx->info);
+
+    peer_log(peer, "waiting-disconnected");
+
+    assert (0 == array_count(ctx->pongCallback));
+    assert (NULL == ctx->mempoolCallback);
```
michael-brd: These probably aren't needed. Added them due to my unfamiliarity with this aspect of the BTC code (assert would catch if we needed to handle this).

bitcoin/BRPeer.c
```diff
@@ -1263,15 +1274,36 @@ void BRPeerDisconnect(BRPeer *peer)
...
 
     if (_peerCheckAndGetSocket(ctx, &socket)) {
         pthread_mutex_lock(&ctx->lock);
         ctx->status = BRPeerStatusDisconnected;
         pthread_mutex_unlock(&ctx->lock);
 
         if (shutdown(socket, SHUT_RDWR) < 0) peer_log(peer, "%s", strerror(errno));
-        close(socket);
     }
+
```
michael-brd: Likely the most controversial part; the spawning of a "cleanup"-like thread. Also, the dual nature of this function...

bitcoin/BRPeer.h
```diff
@@ -93,7 +93,8 @@ extern "C" {
 typedef enum {
     BRPeerStatusDisconnected = 0,
     BRPeerStatusConnecting,
-    BRPeerStatusConnected
+    BRPeerStatusConnected,
+    BRPeerStatusWaiting
```
michael-brd: Roll the separate `waitingForNetwork` flag into this.

bitcoin/BRPeerManager.c
```diff
@@ -839,7 +839,7 @@ static void _peerDisconnected(void *info, int error)
         if (manager->connectFailureCount > MAX_CONNECT_FAILURES) manager->connectFailureCount = MAX_CONNECT_FAILURES;
     }
 
-    if (! manager->isConnected && manager->connectFailureCount == MAX_CONNECT_FAILURES) {
+    if (! manager->isConnected && manager->connectFailureCount >= MAX_CONNECT_FAILURES) {
```
michael-brd: I'd seen cases where this was greater than MAX_CONNECT_FAILURES. In `BRPeerManagerConnect` and `BRPeerManagerPublishTx`, we checked for `>=`, mirrored that here as I had seen cases where it was greater than at this point.

bitcoin/BRPeerManager.c
```diff
@@ -1679,7 +1673,7 @@ void BRPeerManagerDisconnect(BRPeerManager *manager)
         p = manager->connectedPeers[i - 1];
         manager->connectFailureCount = MAX_CONNECT_FAILURES; // prevent futher automatic reconnect attempts
         BRPeerDisconnect(p);
-        if (BRPeerConnectStatus(p) == BRPeerStatusConnecting) manager->peerThreadCount--; // waiting for network
+        while (BRPeerConnectStatus(p) == BRPeerStatusConnecting) BRPeerDisconnect(p);
```
michael-brd: This makes me leery (endless loop anyone?) but the logic *should* be that this only happens when the thread is being spawned. Once spawned, the BRPeer thread will set the status to connected or disconnected. In either case, this will break out (*fingers crossed but also seen in testing*).

bitcoin/testBwm.c
```diff
@@ -1399,7 +1410,7 @@ BRRunTestWalletManagerFileService (const char *storagePath) {
     BRTransactionFree(tx);
     BRTransactionFree(tx2);
     BRSetFree (transactionSet);
-    
+
```
EBGToo: Me again?

michael-brd: I think so (not a big deal). Do you exclusively edit in XCode? Wouldn't be surprised if the XCode 11 update was somehow to blame.

bitcoin/BRPeer.c
yuhyeok2004:
```suggestion
    assert (0 == array_count(ctx->pongCallback));
    assert (NULL == ctx->mempoolCallback);
```

bitcoin/BRPeer.c
yuhyeok2004:
```suggestion

```

yuhyeok2004:
```suggestion

```

yuhyeok2004:
Duplicate of #

PR #297, opened 2019-10-08
CORE-558: In Core Crypto, Migrate Block Blobs w/ Correct Bytes.

EBGToo:
Confirmed w/ Ehsan in a debugging session.


ewerx:
Verified successful migration of BTC + BCH wallets with no address sanitizer errors raised:
`[SYS] migrating btc database: 19 txns / 3795 blocks / 0 peers`
`[SYS] migrating bch database: 5 txns / 951 blocks / 0 peers`

PR #303, opened 2019-10-10
CORE-657: In _BRPeerOpenSocket protect `ctx->socket`

EBGToo:
I checked the call hierarchy to confirm that `_BRPeerOpenSocket` is never called with the `ctx-lock` held.

### DIFF
--- a/bitcoin/BRPeer.c
+++ b/bitcoin/BRPeer.c
@@ -872,24 +872,27 @@ static int _BRPeerOpenSocket(BRPeer *peer, int domain, double timeout, int *erro
     fd_set fds;
     socklen_t addrLen, optLen;
     int count, arg = 0, err = 0, on = 1, r = 1;
+    int sock;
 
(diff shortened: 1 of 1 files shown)


bitcoin/BRPeer.c
```diff
@@ -909,18 +912,18 @@ static int _BRPeerOpenSocket(BRPeer *peer, int domain, double timeout, int *erro
...
         if (err == EINPROGRESS) {
             err = 0;
             optLen = sizeof(err);
             tv.tv_sec = timeout;
             tv.tv_usec = (long)(timeout*1000000) % 1000000;
             FD_ZERO(&fds);
-            FD_SET(ctx->socket, &fds);
-            count = select(ctx->socket + 1, NULL, &fds, NULL, &tv);
+            FD_SET(sock, &fds);
+            count = select(sock + 1, NULL, &fds, NULL, &tv);
```
michael-brd: Just noticed this; its been a while since I've written Berkeley socket code but how is the first argument the socket file descriptor?

PR #304, opened 2019-10-10
CORE-564: In BREventHandler protect `handler->thread` access.

EBGToo:
Maybe sufficient?

### DIFF
--- a/ethereum/event/BREvent.c
+++ b/ethereum/event/BREvent.c
@@ -190,8 +190,10 @@ eventHandlerDestroy (BREventHandler handler) {
     eventHandlerStop(handler);
 
     // ... then kill
+    pthread_mutex_lock (&handler->lock);
     assert (PTHREAD_NULL == handler->thread);
-    pthread_mutex_destroy(&handler->lock);
+    pthread_mutex_unlock  (&handler->lock);
+    pthread_mutex_destroy (&handler->lock);
 
     // release memory
     eventQueueDestroy(handler->queue);
@@ -282,14 +284,18 @@ eventHandlerStop (BREventHandler handler) {
 
 extern int
 eventHandlerIsCurrentThread (BREventHandler handler) {
-    // TODO(fix): This is a hack; fix the ordering such that `handler->thread` is
-    //            is properly set by the time `eventHandlerThread()` runs (CORE-564)
-    return PTHREAD_NULL == handler->thread || pthread_self() == handler->thread;
+    pthread_mutex_lock(&handler->lock);
+    int result = (pthread_self() == handler->thread);
+    pthread_mutex_unlock(&handler->lock);
+    return result;
 }
 
 extern int
 eventHandlerIsRunning (BREventHandler handler) {
-    return PTHREAD_NULL != handler->thread;
+    pthread_mutex_lock(&handler->lock);
+    int result = (PTHREAD_NULL != handler->thread);
+    pthread_mutex_unlock(&handler->lock);
+    return result;
 }
 
 extern BREventStatus
--- a/ethereum/event/BREvent.c
+++ b/ethereum/event/BREvent.c
@@ -65,6 +65,9 @@ struct BREventHandlerRecord {
     // The thread handling events.
     pthread_t thread;
 
+    // A cond-var for exiting the thread.
+    pthread_cond_t threadExit;
+
     // A lock on internal state
     pthread_mutex_t lock;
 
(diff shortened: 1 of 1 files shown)


michael-brd:
> Maybe sufficient?

Nope; deadlock.

Screen Shot 2019-10-10 at 3 23 50 PM

EBGToo:
Nice.  So this is the identical situation that `lesStop()` faced and the 'fix' is easy (and logical).

ethereum/event/BREvent.c
```diff
@@ -268,10 +288,16 @@ eventHandlerStop (BREventHandler handler) {
...
 
-        // Wait for the thread.
+        // Wait for the thread.  We must release the lock here because the `eventHandlerThread`
+        // could itself be waiting on the lock (only as part of a running, dispatched function).
+        // Once we give the lock, the dispatched function will complete, the `eventHandlerThread`
+        // will return to the EventQueue and see the above 'Wait Abort' cond-var signaled.  At
+        // that point, `eventHandlerThread` will finish and the following 'join' returns.
+
+        // TODO: Race on 'start/stop' here
+        pthread_mutex_unlock(&handler->lock);
```
michael-brd: Few issues here:
1. Flagged by TSAN as a data race
2. Can be PTHREAD_NULL, as it is set by the other thread
3. Can result in multiple simultaneous calls to pthread_join (from man page: "The results of multiple simultaneous calls to pthread_join() specifying the same target thread are undefined.")

EBGToo:
The red 'change requested' *exactly* represents this issues status. :-)

Pushed again; fixed this time?

ethereum/event/BREvent.c
```diff
@@ -112,6 +115,13 @@ eventHandlerCreate (const char *name,
         pthread_mutexattr_destroy(&attr);
     }
 
+    {
+        pthread_condattr_t attr;
+        pthread_condattr_init(&attr);
+        pthread_cond_init(&handler->cond, &attr);
```
michael-brd: Doesn't build; no field named `cond`.

ethereum/event/BREvent.c
```diff
@@ -190,8 +211,10 @@ eventHandlerDestroy (BREventHandler handler) {
     eventHandlerStop(handler);
 
     // ... then kill
+    pthread_mutex_lock (&handler->lock);
     assert (PTHREAD_NULL == handler->thread);
-    pthread_mutex_destroy(&handler->lock);
+    pthread_mutex_unlock  (&handler->lock);
+    pthread_mutex_destroy (&handler->lock);
```
michael-brd: Missing `pthread_cond_destroy()`?

ethereum/event/BREvent.c
EBGToo: now it does.

ethereum/event/BREvent.c
EBGToo: Yes.

ethereum/event/BREvent.c
```diff
@@ -268,10 +299,17 @@ eventHandlerStop (BREventHandler handler) {
...
+        // could itself be waiting on the lock (only as part of a running, dispatched function).
+        // Once we give the lock, the dispatched function will complete, the `eventHandlerThread`
+        // will return to the EventQueue and see the above 'Wait Abort' cond-var signaled.  At
+        // that point, `eventHandlerThread` will finish and signal 'threadExit`.  The following
+        // cond_wait will complete and reaquire the lock.
+        //
+        // Note, while we wait we will not hold the lock but handle->thread will be non-NULL; so,
+        // we don't have a start/stop race.
+        //
+        pthread_cond_wait (&handler->threadExit, &handler->lock);
```
michael-brd: Think a `pthread_join` is required to clean up thread resources, no?

EBGToo: Yes

ethereum/event/BREvent.c
```diff
@@ -268,11 +302,19 @@ eventHandlerStop (BREventHandler handler) {
...
```
michael-brd: Really down the rabbit hole now (threading sucks).

Since `pthread_cond_wait` drops the lock, could we have multiple callers of eventHandlerStop waiting in the call? Bunch of issues could be unearthed by that (only one is signalled so other block indefinitely, even if all signalled multiple calls to detach, etc.).

Since the `eventHandlerThread` routine clears the `handler->thread` to `PTHREAD_NULL`, I think an `eventHandlerStart` call could execute before the `pthread_cond_wait` wakes up. That might lead to logic issues with dropped events/abort resets?

Lastly, probably unrealistic but if `eventHandlerThread` ever finished “early” (i.e. while `eventHandlerStop` wasn’t waiting, like if `eventQueueDequeueWait` returning an error condition), `eventHandlerStop` would block indefinitely, I think?

EBGToo:
My recent changes did nothing materially.  The complexity of Object A coordinating Object A's thread is (inexplicably) difficult.  Closing.  Consider later.

PR #306, opened 2019-10-10
CORE-660: Add ed25529 key generation files

carlcherry:


EBGToo:
I added the files to coreMacOS; ready to go

PR #308, opened 2019-10-11
CORE-658: Add syncRecommended event

michael-brd:
I included a depth in the sync recommended event. Should allow consumers to kick off a sync, if they deem the are in a state to do so, without having to determine where to start. Should give us flexibility in case other blockchains ever wanted to recommend a sync.

### DIFF
--- a/Swift/BRCryptoTests/BRCryptoDebugSupport.swift
+++ b/Swift/BRCryptoTests/BRCryptoDebugSupport.swift
@@ -78,6 +78,8 @@ extension WalletManagerEvent: MatchableEvent {
         case (.syncStarted, .syncStarted): return true
         case (let .syncProgress (progress1), let .syncProgress (progress2)):
             return !strict || (progress1 == progress2)
+        case (let .syncRecommended (depth), let .syncRecommended (depth)):
+            return !strict || (depth == depth)
         case (let .syncEnded (error1), let .syncEnded (error2)):
             return !strict || (error1 == error2)
         case (let .blockUpdated (height1), let .blockUpdated (height2)):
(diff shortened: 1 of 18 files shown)


Swift/BRCryptoTests/BRCryptoDebugSupport.swift
```diff
@@ -78,6 +78,8 @@ extension WalletManagerEvent: MatchableEvent {
         case (.syncStarted, .syncStarted): return true
         case (let .syncProgress (progress1), let .syncProgress (progress2)):
             return !strict || (progress1 == progress2)
+        case (let .syncRecommended (depth), let .syncRecommended (depth)):
+            return !strict || (depth == depth)
```
EBGToo: Needs to be `depth1` and `depth2`

bitcoin/BRWalletManagerEvent.c
```diff
@@ -102,9 +103,10 @@ static BREventType bwmSignalTxDeletedEventType = {
...
 extern void
 bwmSignalTxDeleted (BRWalletManager manager,
-                    UInt256 hash) {
+                    UInt256 hash,
+                    int recommendRescan) {
     BRWalletManagerWalletTxDeletedEvent message =
-    { { NULL, &bwmSignalTxDeletedEventType}, manager, hash};
+    { { NULL, &bwmSignalTxDeletedEventType}, manager, hash, recommendRescan};
     eventHandlerSignalEvent (manager->handler, (BREvent*) &message);
 }
```
EBGToo: Ignoring `notifyUser`?  That is fine (added later if needed).

michael-brd: The `notifyUser` actually isn't tied to the rescan. It seems to be an indicator if if the user should be informed of the deletion (i.e. the logic sometimes has `notifyUser` set but `recommendRescan` not).

In our case, we announce every deletion (and every addition) to the CWM.

Swift/BRCryptoTests/BRCryptoDebugSupport.swift
michael-brd: Nice catch!

PR #309, opened 2019-10-11
CORE-662: Add runCryptoTestsWithAccountAndNetwork to Swift tests

michael-brd:


michael-brd:
@EBGToo, these tests (crypto ones) currently fail due to cryptoWalletManagerConnect() causing a disconnect/re-connect each time in BTC:P2P. This is because of we set the peer each time and the P2P code isn't smart enough to notice that the peer hasn't changed.

What do we think about updating the P2P code to *not* disconnect unless the fixed peer has changed?

Alternatively, we can update the tests, if we like that behaviour instead.

michael-brd:
> @EBGToo, these tests (crypto ones) currently fail due to cryptoWalletManagerConnect() causing a disconnect/re-connect each time in BTC:P2P. This is because of we set the peer each time and the P2P code isn't smart enough to notice that the peer hasn't changed.
> 
> What do we think about updating the P2P code to _not_ disconnect unless the fixed peer has changed?
> 
> Alternatively, we can update the tests, if we like that behaviour instead.

The change could be something as simple as:
```
diff --git a/bitcoin/BRPeerManager.c b/bitcoin/BRPeerManager.c
index 97684029..29fb756d 100644
--- a/bitcoin/BRPeerManager.c
+++ b/bitcoin/BRPeerManager.c
@@ -1555,12 +1555,20 @@ void BRPeerManagerSetCallbacks(BRPeerManager *manager, void *info,
 void BRPeerManagerSetFixedPeer(BRPeerManager *manager, UInt128 address, uint16_t port)
 {
     assert(manager != NULL);
-    BRPeerManagerDisconnect(manager);
+
     pthread_mutex_lock(&manager->lock);
-    manager->maxConnectCount = UInt128IsZero(address) ? PEER_MAX_CONNECTIONS : 1;
-    manager->fixedPeer = ((const BRPeer) { address, port, 0, 0, 0 });
-    array_clear(manager->peers);
+    int samePeer = (UInt128Eq(address, manager->fixedPeer.address) &&
+                    (port == manager->fixedPeer.port || UInt128IsZero(address)));
     pthread_mutex_unlock(&manager->lock);
+
+    if (!samePeer) {
+        BRPeerManagerDisconnect(manager);
+        pthread_mutex_lock(&manager->lock);
+        manager->maxConnectCount = UInt128IsZero(address) ? PEER_MAX_CONNECTIONS : 1;
+        manager->fixedPeer = ((const BRPeer) { address, port, 0, 0, 0 });
+        array_clear(manager->peers);
+        pthread_mutex_unlock(&manager->lock);
+    }
 }
```

Swift/CoreTests/CoreTests-Bridging-Header.h
```diff
@@ -13,6 +13,12 @@
 //  Use this file to import your target's public headers that you would like to expose to Swift.
 //
 #include "ethereum/BREthereum.h"
+#include "crypto/BRCryptoAccount.h"
+#include "crypto/BRCryptoAmount.h"
+#include "crypto/BRCryptoCurrency.h"
+#include "crypto/BRCryptoNetwork.h"
+#include "crypto/BRCryptoPrivate.h"
+#include "crypto/BRCryptoUnit.h"
```
EBGToo: This is a preemptive listing of headers?  Only 'Network' and 'Account' are actually needed?

crypto/BRCryptoBase.h
```diff
@@ -97,8 +97,12 @@ extern "C" {
...
 
 #if !defined (private_extern)
-#  define private_extern          extern
-#endif
+#  if defined (DEBUG)
+#    define private_extern
+#  else
+#    define private_extern          extern
+#  endif
+#endif /* !defined (private_extern) */
```
EBGToo: Sorry, what is this doing?

Swift/CoreTests/CoreTests-Bridging-Header.h
michael-brd: The other ones are used to flush out a network (base unit, currency, etc.).

crypto/BRCryptoBase.h
michael-brd: Exposes `private_extern` methods to the tests. Unresolved symbols without this.

EBGToo:
Do we have a JIRA for the 'set peer' issue?

michael-brd:
> Do we have a JIRA for the 'set peer' issue?

Not as yet; not clear if there was an issue to fix/what it is. From your point of view, do we want to fix the "connect always disconnects due to the set fixed peer call" or do you want the tests to account for that?

EBGToo:
> > Do we have a JIRA for the 'set peer' issue?
> 
> Not as yet; not clear if there was an issue to fix/what it is. From your point of view, do we want to fix the "connect always disconnects due to the set fixed peer call" or do you want the tests to account for that?

I don't think we should disconnect/connect w/ an unchaining peer.

PR #312, opened 2019-10-15
CORE-666: Check if fixed peer has changed in BRPeerManagerSetFixedPeer

michael-brd:


bitcoin/BRPeerManager.c
```diff
@@ -1555,12 +1555,20 @@ void BRPeerManagerSetCallbacks(BRPeerManager *manager, void *info,
...
 {
     assert(manager != NULL);
-    BRPeerManagerDisconnect(manager);
+
     pthread_mutex_lock(&manager->lock);
-    manager->maxConnectCount = UInt128IsZero(address) ? PEER_MAX_CONNECTIONS : 1;
-    manager->fixedPeer = ((const BRPeer) { address, port, 0, 0, 0 });
-    array_clear(manager->peers);
+    int samePeer = (UInt128Eq(address, manager->fixedPeer.address) &&
+                    (port == manager->fixedPeer.port || UInt128IsZero(address)));
```
EBGToo: Why is the 'IsZero' check part of the 'port' logic?  Hard to follow the logic - same peer if: addresses match, ports don't match, address is zero?

michael-brd: The check is:
* address match AND
* ports match OR address is zero (i.e. peer is being cleared, thus port doesn't matter)

EBGToo: Thanks for this detail:
```
bool samePeer = (
    // same peer if addresses match and ports match
    (address == manager->fixedPeer.address && port == manager->fixedPeer.port) ||
    // OR same peer if addresses match and the address is zero (and thus port is irrelevant)
    (address == manager->fixedPeer.address && address == 0)
);
```

PR #315, opened 2019-10-17
CORE-668: Add asBlockChainDBModelCurrency in Java Core Crypto

michael-brd:


EBGToo:
I pushed a change to the 'code' case for currency and currency denomination.  For Swift.  If I shouldn't ursurp this PR, for Java, with a Swift change, we can unwind.  A change to Java is implied with this Swift change.

Swift/BRCrypto/BRCryptoSystem.swift
```diff
@@ -103,10 +103,10 @@ public final class System {
                          (name: "Gwei",  code: "gwei", decimals:  9, symbol: BlockChainDB.Model.lookupSymbol ("gwei")),
                          (name: "Ether", code: "eth",  decimals: 18, symbol: BlockChainDB.Model.lookupSymbol ("eth"))]),
 
-        (id: System.makeCurrencyIdentifierERC20 ("ethereum-mainnet", BlockChainDB.Model.addressBRDMainnet), name: "BRD Token", code: "BRD", type: "erc20", blockchainID: "ethereum-mainnet",
+        (id: System.makeCurrencyIdentifierERC20 ("ethereum-mainnet", BlockChainDB.Model.addressBRDMainnet), name: "BRD Token", code: "brd", type: "erc20", blockchainID: "ethereum-mainnet",
          address: BlockChainDB.Model.addressBRDMainnet, verified: true,
-         demoninations: [(name: "BRD Token INT", code: "BRDI",  decimals:  0, symbol: "brdi"),
-                         (name: "BRD Token",     code: "BRD",   decimals: 18, symbol: "brd")]),
+         demoninations: [(name: "BRD Token INT", code: "brdi",  decimals:  0, symbol: "brdi"),
```
michael-brd: Any issue with this (code formatting) diverging from `currencyToDefaultBaseUnit` and `asBlockChainDBModelCurrency` behaviour?

michael-brd:
> I pushed a change to the 'code' case for currency and currency denomination. For Swift. If I shouldn't ursurp this PR, for Java, with a Swift change, we can unwind. A change to Java is implied with this Swift change.

Made the Java change. Have an open question to you but otherwise good to go.

Swift/BRCrypto/BRCryptoSystem.swift
EBGToo: Looks like it.  This:
```
                let code         = code.uppercased()
```
should be 'lowercased()'?

michael-brd: > Looks like it. This:
> 
> ```
>                 let code         = code.uppercased()
> ```
> 
> should be 'lowercased()'?

Can you update the Swift? I'll mirror in Java. You have the "mental model" here as to how these should be.

EBGToo: And these:
```
                            (name: "\(code)_INTEGER",
                                code: "\(code)_INTEGER",
                                decimals: 0,
                                symbol: "\(code)I"),   // BRDI -> BaseUnit

```
should have the code and symbol be "\(code)i" ?

EBGToo:
Let me update the Swift.

EBGToo:
> Should code be lowercased here?

Still I missed something?

michael-brd:
> > Should code be lowercased here?
> 
> Still I missed something?

I'm not sure :), that's the question. What case is code supposed to be? `asBlockChainDBModelCurrency` forces it to uppercase.

Java/CoreCrypto/src/main/java/com/breadwallet/corecrypto/Blockchains.java
```diff
@@ -250,19 +206,41 @@
...
+
+    /* package */
+    static List<CurrencyDenomination> makeCurrencyDemominationsErc20 (String code, UnsignedInteger decimals) {
+        String name = code.toUpperCase(Locale.ROOT);
+        code = code.toLowerCase(Locale.ROOT);
+
+        return ImmutableList.of(
+                new CurrencyDenomination(
+                        String.format(Locale.ROOT, "%s Token INT", name),
+                        String.format(Locale.ROOT, "%si", code),
```
EBGToo: I don't think these are always ERC20 tokens (although in practice I think they are).

Java/CoreCrypto/src/main/java/com/breadwallet/corecrypto/System.java
```diff
@@ -226,6 +227,36 @@ static System create(ScheduledExecutorService executor,
...
+                                                                                           String code,
+                                                                                           String type,
+                                                                                           UnsignedInteger decimals) {
+        int index = uids.indexOf(':');
+        if (index == -1) return Optional.absent();
+
+        type = type.toLowerCase(Locale.ROOT);
+        if (!"erc20".equals(type) && !"native".equals(type)) return Optional.absent();
+
+        code = code.toUpperCase(Locale.ROOT);
```
EBGToo: No lower-cased.

Java/CoreCrypto/src/main/java/com/breadwallet/corecrypto/Blockchains.java
michael-brd: I don't follow the comment here; isn't this an exact replica of `makeCurrencyDemominationsERC20` that you just added in Swift?

PR #320, opened 2019-10-22
CORE-684: In Core Ripple, Define BRRippleAddress.[ch]

carlcherry:
Added the BRRippleAddress files and functions
Modified the other classes to use address functions
Modified the unit test
**NOTE:** whenever BRRippleAddress objects are going "in" or "out" of one of our objects (wallet, transfer, etc) they are copied.  The caller always continues to own any BRRippleAddress they create or ask for so it is clear they must clean them up.
**NOTE:** did not change anything in the generic folder


EBGToo:
@carlcherry In the end I updated BRGenericRipple.c - to keep compilation working - even though I'll rework some stuff in that file.

PR #332, opened 2019-10-24
CORE-699: Add cryptoTransferGetAmountDirectedNet

EBGToo:
Implement `cryptoWalletGenerateBalance()` - used exclusively for a GEN wallet’s balance - with `cryptoTransferGetAmountDirectedNet()`.

### DIFF
--- a/crypto/BRCryptoTransfer.c
+++ b/crypto/BRCryptoTransfer.c
@@ -393,6 +393,33 @@ cryptoTransferGetAmountDirected (BRCryptoTransfer transfer) {
     return amount;
 }
 
+extern BRCryptoAmount
+cryptoTransferGetAmountDirectedNet (BRCryptoTransfer transfer) {
+    BRCryptoAmount amount = cryptoTransferGetAmountDirected (transfer);
+
+    // If the transfer->unit and transfer->unitForFee differ then there is no fee
+    if (cryptoUnitIsIdentical (transfer->unit, transfer->unitForFee))
+        return amount;
+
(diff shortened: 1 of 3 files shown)


EBGToo:
It is only used for GEN because BTC and ETH each have their own wallets that keep track of the balance (for BTC by summing UTXOs; for ETH based on an 'account state' query).

crypto/BRCryptoTransfer.c
```diff
@@ -393,6 +393,33 @@ cryptoTransferGetAmountDirected (BRCryptoTransfer transfer) {
...
 
+extern BRCryptoAmount
+cryptoTransferGetAmountDirectedNet (BRCryptoTransfer transfer) {
+    BRCryptoAmount amount = cryptoTransferGetAmountDirected (transfer);
+
+    // If the transfer->unit and transfer->unitForFee differ then there is no fee
+    if (cryptoUnitIsIdentical (transfer->unit, transfer->unitForFee))
+        return amount;
+
+    BRCryptoFeeBasis feeBasis = (NULL != transfer->feeBasisConfirmed
```
michael-brd: TSAN issue here, likely.

EBGToo: This would be the same TSAN issue suffered by: `cryptoTransferGetAmountDirected()`, `cryptoTransferGetAmount()`, etc?  That is, the issue is not specific to `cryptoTransferGetAmountDirectedNet()`?

EBGToo: Is so, I'll add a TSAN issue.

EBGToo: I updated CORE-462

PR #333, opened 2019-10-25
CORE-652: GEN Transfer Reader/Writer for FileService

EBGToo:


generic/BRGeneric.c
```diff
@@ -204,6 +209,23 @@ genTransferSerialize (BRGenericTransfer transfer, size_t *bytesCount) {
     return transfer->handlers.getSerialization (transfer->ref, bytesCount);
 }
 
+static size_t
+genTransferSetHash (const void *transferPtr) {
+    BRGenericTransfer transfer = (BRGenericTransfer) transferPtr;
+    return genTransferGetHash (transfer).value.u32[0];
```
carlcherry: Isn't this suppose to be "SetHash"

EBGToo: It is actually correct - but about the worst possible naming...

EBGToo: Addressed issue, renamed functions.

EBGToo:
@carlcherry This is ready for review (moved from 'draft')

PR #336, opened 2019-10-28
CORE-642: In Core Generic, add genWalletAddTransfer()

EBGToo:
@carlcherry Can you explore this change with your Ripple BDB?


generic/BRGenericRipple.c
```diff
@@ -73,8 +71,6 @@ genericRippleAccountSignTransferWithKey (BRGenericAccountRef account,
     // Get the transaction pointer from this transfer
     BRRippleTransaction transaction = rippleTransferGetTransaction ((BRRippleTransfer) transfer);
     if (transaction) {
-        // Hard code the sequence to 7
-        rippleAccountSetSequence ((BRRippleAccount) account, 7);
 //        rippleAccountSignTransaction(account, transaction, seed);
         assert (0);
```
carlcherry: Should we just remove the function instead?

EBGToo: We could.

EBGToo:
Also, if a Ripple Wallet now holds all the transfers, can the Ripple Wallet compute a correct balance?  If so, the cryptoWalletUpdateBalance() function, calling cryptoTransferGetAmountDirectedNet(), is no longer needed.

EBGToo:
@carlcherry I updated the Ripple wallet balance to include the transfer fee.

PR #340, opened 2019-10-30
CORE-711: In BRCryptoTransferState use BRCryptoFeeBasis

EBGToo:
'Draft' waiting on CORE-696.


Swift/BRCrypto/BRCryptoTransfer.swift
```diff
@@ -321,7 +321,9 @@ public enum TransferState {
             confirmation: TransferConfirmation (blockNumber: core.u.included.blockNumber,
                                                 transactionIndex: core.u.included.transactionIndex,
                                                 timestamp: core.u.included.timestamp,
-                                                fee: core.u.included.fee.map { Amount (core: $0, take: false) }))
+                                                fee: core.u.included.feeBasis
```
michael-brd: Leaking `feeBasis` here?

EBGToo: Is it?  The caller to `init` owns `core` and thus `feeBasis`.  The `core.u.included.feeBasis` doesn't take the reference.  Am I missing it?

michael-brd: The C code has taken a reference to the fee basis before passing it up to the Swift/Java layer. You then get the fee out of it but you don't give the feebasis reference up (that I notice, maybe I am missing it?).

EBGToo: Addressed w/ cryptoTransferStateRelease() in the TransferState initializer.  Does that appear sufficient?

crypto/BRCryptoTransfer.c
```diff
@@ -811,7 +811,7 @@ cryptoTransferStateRelease (BRCryptoTransferState *state) {
     switch (state->type) {
         case CRYPTO_TRANSFER_STATE_INCLUDED: {
             if (NULL != state->u.included.feeBasis) {
-                cryptoFeeBasisTake (state->u.included.feeBasis);
+                cryptoFeeBasisGive (state->u.included.feeBasis);
```
michael-brd: Good catch.

PR #341, opened 2019-10-30
CORE-712: In Core Hedera: Add "node account id" into serialized bytes

carlcherry:
When serializing a transaction - prepend 24 bytes to hold the
shard, realm, and accountid of the expected Hedera submission node

The Hedera java SDK being used on the server does not have any public function to get at the account node id from the serialized bytes - so we need to send them up.

### DIFF
--- a/hedera/BRHederaTransaction.c
+++ b/hedera/BRHederaTransaction.c
@@ -109,15 +109,29 @@ hederaTransactionSignTransaction (BRHederaTransaction transaction,
     ed25519_sign(signature, body, bodySize, publicKey.pubKey, privateKey);
 
     // Serialize the full transaction including signature and public key
-    transaction->serializedBytes = hederaTransactionPack (signature, 64,
-                                                          publicKey.pubKey, 32,
-                                                          body, bodySize,
-                                                          &transaction->serializedSize);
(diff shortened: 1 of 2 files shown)


hedera/BRHederaTransaction.c
```diff
@@ -109,15 +109,29 @@ hederaTransactionSignTransaction (BRHederaTransaction transaction,
...
+    // The server needs to know what node we are sending to - so let's add those bytes now.
+    // We will need 24 bytes to hold the 3 numbers - so create a new buffers of that size
+    transaction->serializedBytes = calloc(1, transaction->serializedSize + 24);
+
+    int64_t shard = htonl(hederaAddressGetShard(nodeAddress));
+    memcpy(&transaction->serializedBytes[0], &shard, 8);
+    int64_t realm = htonl(hederaAddressGetRealm(nodeAddress));
+    memcpy(&transaction->serializedBytes[8], &realm, 8);
+    int64_t account = htonl(hederaAddressGetAccount(nodeAddress));
+    memcpy(&transaction->serializedBytes[16], &account, 8);
```
EBGToo: Isn't 24 equal to `3 *sizeof (<something>)` where <something> is a typedef for the Hedera nodeAddress individual components?

PR #345, opened 2019-10-31
CORE-715: Update BlockChainDB Subscription JSON types

EBGToo:
Fill out the unit tests.  Update the JWT token.

This does not address subscriptions in the Crypto interface; only connectivity to the BDB.

Requires Java changes.

### DIFF
--- a/Swift/BRCrypto/common/BRBlockChainDB.swift
+++ b/Swift/BRCrypto/common/BRBlockChainDB.swift
@@ -78,6 +78,7 @@ public class BlockChainDB {
     /// A Subscription specificiation
     internal var subscription: Subscription? = nil
 
+    #if false
     private var modelSubscription: Model.Subscription? {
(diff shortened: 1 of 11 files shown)


Swift/BRCrypto/common/BRBlockChainDB.swift
```diff
@@ -593,47 +665,50 @@ public class BlockChainDB {
...
         }
     }
 
+    public func createSubscription (_ subscription: Model.Subscription, // TODO: Hackily
+                                    completion: @escaping (Result<Model.Subscription, QueryError>) -> Void) {
+        makeSubscriptionRequest (
+            path: "subscriptions",
+            data: [
+                // Can use asJSON(Subscription) because that will include the 'id'
+                "device_id"       : subscription.device,
```
EBGToo: "Can use" ==> "Can not use"

PR #346, opened 2019-10-31
CORE-718: In Core ripple: handle unknown address

carlcherry:
Create specific bytes for the unknown address (or empty string)
Return "unknown" when the addres is our special array of bytes
Add unit tests

### DIFF
--- a/ripple/BRRippleAddress.c
+++ b/ripple/BRRippleAddress.c
@@ -26,9 +26,21 @@ struct BRRippleAddressRecord {
 };
 
 uint8_t feeAddressBytes[20] = {
-    0x42, 0x52, 0x44, 0x5F, 0x5F, 0x66, 0x65, 0x65,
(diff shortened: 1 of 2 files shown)


ripple/BRRippleAddress.c
```diff
@@ -138,9 +181,11 @@ BRRippleAddress rippleAddressStringToAddress(const char* input)
 extern BRRippleAddress
 rippleAddressCreateFromString(const char * rippleAddressString)
 {
-    // 1 special case so far - the __fee__ address.
+    // 2 special case so far - the __fee__ address and "unknown" address
     // See the note in BRRippleAcount.h with respect to the feeAddressBytes
-    if (strcmp(rippleAddressString, "__fee__") == 0) {
+    if (rippleAddressString == NULL || strlen(rippleAddressString) == 0 || strcmp(rippleAddressString, "unknown") == 0) {
+        return rippleAddressCreateUnknownAddress ();
```
EBGToo: You covered all the possible cases!

PR #347, opened 2019-11-01
CORE-721: For a BTC Transfer, use 'direction' to determine the source/target addresses

EBGToo:


EBGToo:
This is reported to fix the original issue.

PR #348, opened 2019-11-01
CORE-720: Add Java-specific BRCoreKey::encryptNative migration function

michael-brd:
For the Android guys, the call looks like (roughly):
```
byte[] ciphertext = system.migrateBRCoreKeyCiphertext(key, nonce, authenticatedData, ciphertext);
```

### DIFF
--- a/crypto/BRCryptoCipher.h
+++ b/crypto/BRCryptoCipher.h
@@ -64,11 +64,6 @@ extern "C" {
                          const uint8_t *ciphertext,
                          size_t ciphertextLen);
 
-    extern size_t
-    cryptoCipherMigrateBRCoreKeyCiphertextLength (BRCryptoCipher cipher,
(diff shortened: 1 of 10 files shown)


crypto/BRCryptoCipher.h
```diff
@@ -54,16 +54,23 @@ extern "C" {
...
-                               const char *ciphertext,
+                               const uint8_t *ciphertext,
                                size_t ciphertextLen);
 
     extern BRCryptoBoolean
     cryptoCipherDecrypt (BRCryptoCipher cipher,
                          uint8_t *plaintext,
                          size_t plaintextLen,
-                         const char *ciphertext,
+                         const uint8_t *ciphertext,
```
michael-brd: Oversight here in initial commit (should have always been `uint8_t *`; copy-pasta from the BRCryptoCoder I suspect).

PR #349, opened 2019-11-01
CORE-717: Remove BTC/ETH includes from public crypto headers

michael-brd:
- Removed include statements for BTC/ETH in public headers (i.e. not BRCryptoXyzPrivate.h headers)
- Moved functions declarations that used BTC/ETH/GEN types to BRCryptoPrivate.h.
- Added include statements for system headers uncovered by cleanup

Potentially controversial change is to the `BRCryptoCWMEthGetBlocksCallback` declaration in `BRCryptoWalletManagerClient.h`. It was of type `BREthereumSyncInterestSet`. To achieve removal of ETH headers there, needed to change to its alias (unsigned int). Alternatively, could add a typedef for it? Or could add the header back?

### DIFF
--- a/crypto/BRCryptoWalletManagerClient.h
+++ b/crypto/BRCryptoWalletManagerClient.h
@@ -18,10 +18,6 @@
 #include "BRCryptoTransfer.h"
 #include "BRCryptoWallet.h"
 
-#include "ethereum/BREthereum.h"
-#include "bitcoin/BRWalletManager.h"
-#include "generic/BRGenericWalletManager.h"
-
 #ifdef __cplusplus
 extern "C" {
 #endif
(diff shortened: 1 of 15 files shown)


crypto/BRCryptoWalletManagerClient.h
```diff
@@ -95,7 +91,7 @@ extern "C" {
                                             OwnershipGiven BRCryptoCWMClientCallbackState callbackState,
                                             OwnershipKept const char *network,
                                             OwnershipKept const char *address, // disappears immediately
-                                            BREthereumSyncInterestSet interests,
+                                            unsigned int interests,
```
EBGToo: That works.

PR #350, opened 2019-11-04
CORE-329: Add feePerKb as an argument to wallet functions

michael-brd:


EBGToo:
@michael-brd Can you add a comment, perhaps in the BRWallet.h file about what problem this set of (about) five functions is addressing?  Essentially, why aren't BRWallet{Get,Set}FeePerKb() sufficient?

EBGToo:
Or perhaps the explanation can be added near BRWallet{Get,Set}FeePerKb().

michael-brd:
> Perhaps the explanation can be added near BRWallet{Get,Set}FeePerKb(). As a precautionary note.

Done.

EBGToo:
@voisine - This PR adds a number of functions to deal with a wallet’s feePerKb value.  We found ourselves needing to specify a feePerKb when creating a transaction but being in a race condition with the peer-to-peer fee update.  Specifically, BlockSet gives us a set of feePerKb values corresponding to inclusion times; the User selects one and we must create a transaction with that feePerKb.  We end up in a race.  We also needed to: save the wallet’s current feePerKb, set the wallet’s feePerKb to the desired value, create the transaction, restore the saved wallet’s value. 

Another issue, although less likely in our IOS/Android apps, would be performing multiple simultaneous transaction creates or fee calculations.  The wallet's feePerKb would end up in a race.

So, we’ve added some functions to take a feePerKb value directly.

PR #361, opened 2019-11-05
CORE-704: On System.systemExtract failure, cryptoWalletManagerGive() CWM

EBGToo:


EBGToo:
> Nice. One thing to keep in mind, that the Swift doesn't currently account for, some events have references in them that need to be released. If we have a "miss" on CWM/WID/TID, those references leak.
> 
> The Java solves this problem by some very verbose code.

Isn't this covering that 'miss':
```
            transferEventCallback: { (context, cwm, wid, tid, event) in
                precondition (nil != context  && nil != cwm && nil != wid && nil != tid)
                defer { cryptoWalletManagerGive(cwm); cryptoWalletGive(wid); cryptoTransferGive(tid) }
                ...
```

michael-brd:
> > Nice. One thing to keep in mind, that the Swift doesn't currently account for, some events have references in them that need to be released. If we have a "miss" on CWM/WID/TID, those references leak.
> > The Java solves this problem by some very verbose code.
> 
> Isn't this covering that 'miss':
> 
> ```
>             transferEventCallback: { (context, cwm, wid, tid, event) in
>                 precondition (nil != context  && nil != cwm && nil != wid && nil != tid)
>                 defer { cryptoWalletManagerGive(cwm); cryptoWalletGive(wid); cryptoTransferGive(tid) }
>                 ...
> ```

No; think about the CRYPTO_WALLET_EVENT_TRANSFER_ADDED event, or CRYPTO_WALLET_EVENT_BALANCE_UPDATED event, or etc.

I'm referring to events where, within the event structure, we have a given reference.

PR #362, opened 2019-11-06
CORE-722: Add `system.wipe()`

EBGToo:
Rather than adding an argument to `cryptoWalletManagerCreate()` of `wipe` which a) changes the interface, b) is only used once in a million, c) needs to be propagated down to the BWM, EWM and GWM so that the proper file can be wiped, I added a new function `cryptoWalletManagerWipe()` which propagates down to BWM, EWM and GWM and is called one time in a million via a function on `System`.

Reasonable?  Would need a Java analogue. 

Example in CoreDemoListener.swift:
```
                if !success {
                    system.wipe (network: network)
                    let successRetry = system.createWalletManager (network: network,
                                                                   mode: mode,
                                                                   addressScheme: scheme,
                                                                   currencies: currencies)
                    if !successRetry { UIApplication.doError(network: network) }
                }
            }		
```

### DIFF
--- a/support/BRFileService.c
+++ b/support/BRFileService.c
@@ -208,6 +208,17 @@ fileServiceCreateReturnError (BRFileService fs,
     return NULL;
 }
 
+static char *
+fileServiceCreateFilePath (const char *basePath,
+                           const char *currency,
+                           const char *network,
+                           const char *filename) {
+    size_t sdbPathLength = strlen (basePath) + 1 + strlen(currency) + 1 + strlen(network) + 1 + strlen (filename) + 1;
+    char   *sdbPath      = malloc (sdbPathLength);
+    sprintf (sdbPath, "%s/%s-%s-%s", basePath, currency, network, filename);
+    return sdbPath;
+}
+
 extern BRFileService
 fileServiceCreate (const char *basePath,
                    const char *currency,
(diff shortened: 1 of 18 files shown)


support/BRFileService.c
```diff
@@ -820,6 +820,19 @@ fileServiceClearAll (BRFileService fs) {
...
+fileServiceWipe (const char *basePath,
+                 const char *currency,
+                 const char *network) {
+
+    // Locate the SQLITE Database
+     size_t sdbPathLength = strlen (basePath) + 1 + strlen(currency) + 1 + strlen(network) + 1 + strlen (FILE_SERVICE_SDB_FILENAME) + 1;
+     char   *sdbPath      = malloc (sdbPathLength);
+     sprintf (sdbPath, "%s/%s-%s-%s", basePath, currency, network, FILE_SERVICE_SDB_FILENAME);
+
+    return 0 == remove (sdbPath) ? 0 : errno;
```
michael-brd: Leak `sdbPath`? Also, should this path building logic be moved to a function (have it twice in this file)?

michael-brd:
Added the Java equivalent.

support/BRFileService.c
EBGToo: Addressed.

PR #364, opened 2019-11-07
CORE-400: Add cryptoTransferCompare

michael-brd:
Couple of points to consider:

- I created one function (cryptoTransferCompare, ala cryptoAmountCompare); do we want/need more?
- Are you OK with the vanilla nature of the name? Since multiple facets are considered, couldn't come up with something else...
- I sorted "ascending", only because logically it was the simplest for me to reason; good with that?
- It reflects the sort algorithm we use both use in our demo apps (with timestamp thrown in); do we care about ordering for other states (besides INCLUDED)?


EBGToo:
I'm good w/ the change as is.

PR #365, opened 2019-11-07
CORE-734: Memory Issues for GEN/Ripple

EBGToo:


crypto/BRCryptoAccount.c
```diff
@@ -223,6 +223,7 @@ cryptoAccountRelease (BRCryptoAccount account) {
     accountFree(account->eth);
     genAccountRelease(account->xrp);
 
+    free (account->uids);
```
michael-brd: 😦 good catch...

PR #369, opened 2019-11-12
CORE-719: In BRGenericTransfer add a UIDS; Add genTransferEqual()

EBGToo:
When matching GEN BDB Transfers work hard to handle the ‘__fee__’.

Requires a Java update to `system.mergeTransfers()` and an argument addition to `cwmAnnounceGetTransferItemGEN()`

### DIFF
--- a/crypto/BRCryptoHash.c
+++ b/crypto/BRCryptoHash.c
@@ -80,18 +80,27 @@ cryptoHashEqual (BRCryptoHash h1, BRCryptoHash h2) {
             : CRYPTO_FALSE);
 }
 
+static char *
+_cryptoHashAddPrefix (char *hash) {
+    char *result = malloc (strlen (hash) + 3);
+    strcpy (result, "0x");
+    strcat (result, hash);
+    free (hash);
+    return result;
+}
+
 extern char *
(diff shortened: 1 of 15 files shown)


crypto/BRCryptoHash.c
```diff
@@ -80,18 +80,27 @@ cryptoHashEqual (BRCryptoHash h1, BRCryptoHash h2) {
             : CRYPTO_FALSE);
 }
 
+static char *
+_cryptoHashAddPrefix (char *hash) {
+    char *result = malloc (strlen (hash) + 3);
+    strcpy (result, "0x");
+    strcat (result, hash);
+    free (hash);
```
michael-brd: This (the free) seems like a scary side-effect to roll into the function.

EBGToo: That is me obsessing on tail calls.  I pushed a change that at least makes it more explicit.

EBGToo:
There is more work to do on this; specifically handing created GEN transfers.

EBGToo:
I updated to handle a 'created transfers' that is then seen from a BDB recovery.  Notably, update the UIDS.

PR #372, opened 2019-11-15
CORE-742: Improve BTC transaction state handling in crypto

michael-brd:
@EBGToo, placing this in DRAFT for now as I test further but wanted early eyes on it and the approach. In API testing, things look good. Just waiting for a sync to confirm P2P behaviour.

### DIFF
--- a/bitcoin/BRWalletManager.c
+++ b/bitcoin/BRWalletManager.c
@@ -399,6 +399,8 @@ BRWalletSweeperValidate (BRWalletSweeper sweeper) {
 /// MARK: - Transaction Tracking
 
 /**
+ * A Word About Ownership
+ *
  * The BRWallet has ownership over its set of transactions. As a result, it is not safe
  * for the BRWalletManager to access a transaction owned by its wrapped BRWallet, as that
  * transaction may be deleted at any time by the wallet.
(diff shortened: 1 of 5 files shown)


michael-brd:
@EBGToo, changed to "ready" as this looks good in testing.

bitcoin/BRWalletManager.c
```diff
@@ -412,6 +414,39 @@ BRWalletSweeperValidate (BRWalletSweeper sweeper) {
  * The key to this design is that the `refedTransaction` is NEVER, EVER dereferenced. It is
  * used solely for lookups based on identify in callbacks from the BRWallet and
  * BRPeerManager (via the BRSyncManager).
+ *
+ *
+ * A Word About State
+ *
```
EBGToo: Nice explanation.

crypto/BRCryptoWalletManagerClient.c
```diff
@@ -698,7 +666,7 @@ cwmTransactionEventAsBTC (BRWalletManagerClientContext context,
 
             BRCryptoTransferState oldState = cryptoTransferGetState (transfer);
 
-            if (CRYPTO_TRANSFER_STATE_INCLUDED == oldState.type &&
+            if (CRYPTO_TRANSFER_STATE_SUBMITTED != oldState.type &&
                 (0 == event.u.updated.timestamp || TX_UNCONFIRMED == event.u.updated.blockHeight)) {
                 // The transfer is not included so set it to the submitted state at this point
                 BRCryptoTransferState newState = cryptoTransferStateInit (CRYPTO_TRANSFER_STATE_SUBMITTED);
```
EBGToo: Might be the comment 'The transfer is not included ...' is inconsistent with the logic.  Maybe a general explanation in front of the `if` - as in "if the transfer is not submitted and is missing values for timestamp and blockHeight, then return it to submitted" ?

crypto/BRCryptoWalletManagerClient.c
```diff
@@ -652,38 +652,6 @@ cwmTransactionEventAsBTC (BRWalletManagerClientContext context,
                 cryptoUnitGive (unit);
            }
 
-            // ... update state to reflect included if the timestamp and block height are already set
-            if (0 != btcTransaction->timestamp && TX_UNCONFIRMED != btcTransaction->blockHeight) {
-
```
EBGToo: Maybe a comment in the 'empty code' to the effect: "We won't announce a state change because ..."

PR #376, opened 2019-11-20
CORE-747: Convert peer array to pointers before fileServiceReplace call

michael-brd:


leonmendezarrona (commented):
Se pudo?

leonmendezarrona:
Perfecto

PR #378, opened 2019-11-21
CORE-740: Add estimateLimit{Maximum,Minimum}

EBGToo:
This requires Java changes:
- Add Amount.isZero
- Add Wallet estimateLimit(), estimateLimitMaximum(), estimateLimitMinimum()

Note: still performing numeric checks on the various currencies.  Specifically the BTC min value *might not* account for fees.  And I need to confirm C memory issues.  But getting this out there for review.


Swift/BRCrypto/BRCryptoWallet.swift
```diff
@@ -169,18 +169,252 @@ public final class Wallet: Equatable {
...
+            else {
+                // This is extraneous as `cryptoWalletEstimateLimit()` always returns an amount
+                completion (Result.failure (LimitEstimationError.insufficientFunds))
+                return;
+        }
+
+        // If we don't need an estimate, then we invoke `completion` and skip out immediately.  But
+        // include a check on a zero amount - which indicates insufficient funds.
+        if CRYPTO_FALSE == needFeeEstimate {
+            system.queue.async {
```
michael-brd: Why is this done on the async queue but not the "extraneous" call above?

Swift/BRCrypto/BRCryptoWallet.swift
```diff
@@ -169,18 +169,252 @@ public final class Wallet: Equatable {
...
+        // The currency for the fee
+        let currencyForFee = fee.pricePerCostFactor.currency
+
+        guard let walletForFee = self.manager.wallets
+            .first (where: { $0.currency == currencyForFee })
+            else { completion (Result.failure (LimitEstimationError.serviceError)); return }
+
+        // Skip out immediately if we've no balance.
+        if walletForFee.balance.isZero {
+            completion (Result.failure (Wallet.LimitEstimationError.insufficientFunds))
```
michael-brd: Missing return after this?

Swift/BRCrypto/BRCryptoWallet.swift
```diff
@@ -169,18 +169,252 @@ public final class Wallet: Equatable {
...
+            }
+            return
+        }
+
+        // We need an estimate of the fees.
+
+        // The currency for the fee
+        let currencyForFee = fee.pricePerCostFactor.currency
+
+        guard let walletForFee = self.manager.wallets
```
michael-brd: This sort of speaks to this function hanging off the manager, no?

Swift/BRCrypto/BRCryptoWallet.swift
```diff
@@ -169,18 +169,252 @@ public final class Wallet: Equatable {
...
+        case serviceUnavailable
+        case serviceError
+        case insufficientFunds
+
+        static func fromStatus (_ status: BRCryptoStatus) -> LimitEstimationError {
+            switch status {
+            case CRYPTO_ERROR_FAILED: return .serviceError
+            default: return .serviceError // preconditionFailure ("Unknown FeeEstimateError")
+            }
+        }
```
michael-brd: Is this used anywhere?

Swift/BRCrypto/BRCryptoWallet.swift
EBGToo: They all should.  Will update.

Swift/BRCrypto/BRCryptoWallet.swift
EBGToo: yes.

Swift/BRCrypto/BRCryptoWallet.swift
EBGToo: I'd like to move to all 'active' functions to be at WalletManager or System level.  Maybe this is the place to start.

EBGToo: Will defer.

Swift/BRCrypto/BRCryptoWallet.swift
EBGToo: I don't think so.

PR #380, opened 2019-11-26
CORE-750: Public BRCrypto Files into include/

EBGToo:
This moves towards: A SwiftPM distribution and a clear WalletKit public interface (the stuff in include/)

Includes an update to build.gralde to find .../include

Changed the module.modulemap file to not `export *`; which excludes recursive includes and thus C standard libraries are not included.  This means Swift calls to `free` and use of `size_t` are unavailable.  Added some wrappers to BRCrypto (cryptoMemoryFree and BRCryptoCount).

Also removed BRCryptoPrivate.h from the module.modulemap for framework 'Crypto'; added Crypto.Impl which only includes BRCryptoPrivate.h.  Modified swift files to use Crypto.Impl as appropriate.  Expect Crypto.Impl to disappear at some point.

Made a number of cryptoXYXCreate() functions public - but only those that are not BTC, ETH or GEN specific.  For example cryptoUnitCreate().  Use of these functions, via Swift or Java, does not change; but a review of the new public functions will need to occur for general C use.  (For example, are the public function implementations thread safe.)

Some files outside of include/ are still included: BRSyncMode.h, BRInt.h, others.  Expect this change at some point.


EBGToo:
No changes to functionality.

michael-brd:
The corenative-android build failed. I updated the CMakeLists.txt file to get it building again. The JRE builds use the corenative-jre gradle file but the corenative-android build uses the CMakeLists.txt file specified in its gradle file.

PR #385, opened 2019-12-03
CORE-760: Support API_ONLY mode for BCH

EBGToo:


bitcoin/BRSyncManager.c
```diff
@@ -1211,6 +1214,44 @@ BRClientSyncManagerAnnounceGetTransactionsItem (BRClientSyncManager manager,
     }
 }
 
+static BRArrayOf(char *)
+BRClientSyncManagerConvertAddressToString (BRClientSyncManager manager,
+                                           OwnershipGiven BRArrayOf(BRAddress *) addresses) {
+    if (NULL == addresses || 0 == array_count(addresses))
```
michael-brd: leak if count is 0?

EBGToo: Yes.

bitcoin/BRSyncManager.c
```diff
@@ -2041,8 +2087,14 @@ static BRAddress *
 _getWalletAddresses (BRWallet *wallet,size_t *addressCount) {
     assert (addressCount);
 
+    // For BTC, `BRWalletAllAddrs() produces an array of SEGWIT addresses; for BCH, it
+    // produces an array of LEGACY addresses.  See `BRAddressFromHash160()`
     size_t addrCount = BRWalletAllAddrs (wallet, NULL, 0);
 
+    // We'll double up the number of addresses.  For BTC: SEGWIT + LEGACY; for BCH: LEGACY +
```
michael-brd: Is it too much trouble to propagate if it is BCH/BTC to this point?

EBGToo: I thought about it but biased towards minimal changes - once I realized BRSet would handle it.  Better to just pass an argument through?

EBGToo: Maybe better to handle it a different way...  Don't add the LEGACY address here; instead when converting to string, for BTC, double up with SEGWIT + LEGACY? That then leaves BRWalletAllAddrs() as the set of addresses and leaves the string conversion for BTC and BCH in one place.

michael-brd: I like that.

PR #401, opened 2019-12-19
CORE-780: Fix cryptoNetworkInstallBuiltins Handling Non-Native Units

EBGToo:


crypto/BRCryptoNetwork.c
```diff
@@ -775,6 +773,9 @@ cryptoNetworkInstallBuiltins (BRCryptoCount *networksCount) {
                                                  currencySpec->type,
                                                  currencySpec->address);
 
+                BRCryptoUnit unitBase    = NULL;
+                BRCryptoUnit unitDefault = NULL;
```
michael-brd: Is this leaked?

crypto/BRCryptoNetwork.c
```diff
@@ -814,15 +815,16 @@ cryptoNetworkInstallBuiltins (BRCryptoCount *networksCount) {
         }
 
         // Create the Network Fees
+        BRCryptoUnit feeUnit = cryptoNetworkGetUnitAsBase (network, cryptoNetworkGetCurrency(network));
```
michael-brd: Is feeUnit leaked? Is `cryptoNetworkGetCurrency()` result leaked?

EBGToo:
I think I got them all handled.

crypto/BRCryptoNetwork.c
michael-brd: Isn't the cryptoNetworkGetCurrency result leaked?

EBGToo: Fixed that.  And another once over.

michael-brd: The joys of memory management; looks good!

PR #403, opened 2019-12-20
CORE-781: Remove Framework BRCrypto.Impl and BRCryptoPrivate.h

EBGToo:


EBGToo:
Updated CMakeLists.txt; seems no other changes needed (Demo and Unit Test run/succeed)

michael-brd:
> Updated CMakeLists.txt; seems no other changes needed (Demo and Unit Test run/succeed)

Was referring to coerceStringPrefaced removal from java as well.

PR #405, opened 2019-12-20
CORE-759: Remove BRInt.h from the Crypto Public Interface

EBGToo:
As a 'draft'.


Java/CoreNative/src/main/java/com/breadwallet/corenative/support/BRCryptoData16.java
```diff
@@ -1,47 +1,52 @@
...
 
+
 import com.sun.jna.Pointer;
 import com.sun.jna.Structure;
 
 import java.util.Arrays;
 import java.util.List;
 
-public class UInt512 extends Structure {
+public class BRCryptoData16 extends Structure {
```
michael-brd: You didn't like it in bits?

EBGToo: Actually, that was GIT corresponding two files - that aren't really related.  I deleted UInt512.
I added BRCryptoData16 (because UInt128 was in the BRCryptoPeer.h interface).

EBGToo: I took the cues from bitcoin, `uint8_t *md32`, ...

michael-brd: I'm good with it; just picking your brain on thought process.

PR #407, opened 2020-01-02
CORE-787: In Generic Core System.configure(), bias to `supportedNetworks`.

EBGToo:
Discovered this while adding XRP.  Core had it as a builtin but BlockSet did not; we want all builtin blockchains to appear.  (They can be used in P2P mode but will fail if the API endpoint is dead).


EBGToo:
This was tested with the changes upcoming for CORE-786

PR #410, opened 2020-01-03
CORE-773: Pass BlockChainDB transaction status to CWM.

EBGToo:
This doesn't actually use the status.  Just passes it out of 'BlockchainDB' into CWM.


crypto/BRCryptoWalletManagerClient.c
```diff
@@ -1921,6 +1921,7 @@ cwmAnnounceGetBlockNumberFailure (OwnershipKept BRCryptoWalletManager cwm,
 extern void
 cwmAnnounceGetTransactionsItemBTC (OwnershipKept BRCryptoWalletManager cwm,
                                    OwnershipGiven BRCryptoCWMClientCallbackState callbackState,
+                                   BRCryptoTransferStateType status,
```
michael-brd: Future PR to make use of this?

PR #416, opened 2020-01-09
CORE-793: Address Memory Leak Issues

EBGToo:
These leaks were found when displaying and submitting XRP transfers but some of them applied outside of XRP.

I'll run some BTC, etc tests and perhaps add to this PR.

### DIFF
--- a/generic/BRGenericRipple.c
+++ b/generic/BRGenericRipple.c
@@ -212,12 +212,16 @@ genericRippleWalletCreateTransfer (BRGenericWalletRef wallet,
                                    BRGenericAddressRef target,
                                    UInt256 amount,
                                    BRGenericFeeBasis estimatedFeeBasis) {
-    BRRippleAddress source = rippleWalletGetSourceAddress ((BRRippleWallet) wallet);
-    BRRippleUnitDrops drops  = amount.u64[0];
+    BRRippleAddress source  = rippleWalletGetSourceAddress ((BRRippleWallet) wallet);
+    BRRippleUnitDrops drops = amount.u64[0];
 
-    return (BRGenericTransferRef) rippleTransferCreateNew (source,
-                                                           (BRRippleAddress) target,
-                                                           drops);
+    BRRippleTransfer transfer = rippleTransferCreateNew (source,
+                                                         (BRRippleAddress) target,
+                                                         drops);
+
+    rippleAddressFree(source);
+
+    return (BRGenericTransferRef) transfer;
 }
 
 static BRGenericFeeBasis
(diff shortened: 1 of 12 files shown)


EBGToo:
@carlcherry If you didn't look at it carefully, can you confirm the 'ripple' changes in detail?

ripple/BRRippleTransaction.c
```diff
@@ -318,10 +318,11 @@ rippleTransactionSerializeAndSign(BRRippleTransaction transaction, BRKey * priva
     // Serialize and sign the tx bytes
     BRRippleSerializedTransaction serializedBytes = rippleTransactionSerializeImpl (transaction, 0, 0);
     BRRippleSignature sig = signBytes(privateKey, serializedBytes->buffer, serializedBytes->size);
-    free(serializedBytes); // Free the serialized bytes since we have to reserialize with signature below
```
carlcherry: I guess I missed the fact that this structure had a member that was also allocated memory

carlcherry: I know this is not part of your change - but I do notice that down on line 335 we are dereferencing transaction->signedBytes without checking if the pointer is valid

ripple/BRRippleTransaction.c
```diff
@@ -318,10 +318,11 @@ rippleTransactionSerializeAndSign(BRRippleTransaction transaction, BRKey * priva
...
 
     transaction->signedBytes = rippleTransactionSerializeImpl(transaction, sig->signature, sig->sig_length);
-
+    
+    rippleSignatureDelete(sig);
```
carlcherry: Yes - forgot to free this.

generic/BRGenericRipple.c
```diff
@@ -212,12 +212,16 @@ genericRippleWalletCreateTransfer (BRGenericWalletRef wallet,
...
+    BRRippleUnitDrops drops = amount.u64[0];
 
-    return (BRGenericTransferRef) rippleTransferCreateNew (source,
-                                                           (BRRippleAddress) target,
-                                                           drops);
+    BRRippleTransfer transfer = rippleTransferCreateNew (source,
+                                                         (BRRippleAddress) target,
+                                                         drops);
+
+    rippleAddressFree(source);
```
carlcherry: Yes - this is needed.

ripple/BRRippleTransaction.c
EBGToo: Yes, that needs to be fixed!

EBGToo: Like this `return (NULL == transaction->signedBytes ? 0 : transaction->signedBytes->size);`  Okay?

PR #418, opened 2020-01-13
CORE-796: Updates for Core 5.0.0 Release

EBGToo:


EBGToo:
@voisine Care to check the BTC checkpoint additions?

PR #425, opened 2020-01-16
CORE-802: Add genWalletRemTransfer

EBGToo:


generic/BRGeneric.c
```diff
@@ -360,7 +366,7 @@ genWalletCreateTransfer (BRGenericWallet wallet,
                               ? GENERIC_TRANSFER_SENT
                               : GENERIC_TRANSFER_RECEIVED));
 
-    genWalletAddTransfer (wallet, transfer);
+    // We don't add `transfer` to `wallet`.  Let an explicit `genWalletAddTransfer` do that.
```
carlcherry: I think I get it - we don't add it when we create the transfer - it get's created once we pull it down from blockset?  Or when we get a response from the submit?

crypto/BRCryptoWalletManager.c
```diff
@@ -945,6 +965,11 @@ cryptoWalletManagerSubmitSigned (BRCryptoWalletManager cwm,
             // to wallet.  So, we'll add transfer here...
             cryptoWalletAddTransfer (wallet, transfer);
 
+            // Add the signed/submitted transfer to the GEN wallet.  If the submission fails
+            // we'll remove it then.  For now, include transfer when computing the balance and
+            // the sequence-number 
+            genWalletAddTransfer (cryptoWalletAsGEN(wallet), cryptoTransferAsGEN(transfer));
```
carlcherry: OK - so it gets added once it is signed so we have the hash.  Get's removed if a failure happens.

generic/BRGeneric.c
EBGToo: Yes.

crypto/BRCryptoWalletManager.c
EBGToo: Yes

PR #427, opened 2020-01-16
CORE-801: For GEN, generate a TransferEvent.created

EBGToo:
I created a Java JIRA for this.  There is another change to `createTransfer` coming; will handle everything in that ticket.


EBGToo:
> Missing Java change; is that in the followup JIRA?

Yes.

PR #428, opened 2020-01-17
CORE-808: Increment the XRP sequence number when signing.

EBGToo:


ripple/BRRippleAccount.c
```diff
@@ -210,7 +210,8 @@ rippleAccountSignTransaction(BRRippleAccount account, BRRippleTransaction transa
 
     size_t tx_size =
         rippleTransactionSerializeAndSign(transaction, &key, &account->publicKey,
-                                          account->sequence, account->lastLedgerSequence);
+                                          account->sequence + 1,  // next sequence number
```
carlcherry: I think this makes sense - the ripple wallet would always start with a sequence of 0 - the first send would then use a sequence of 1 (and so on.).  When you remove the transfers it fixes things up.  I just have one question about the remove transfer logic.  Does it get removed from the crypto wallet if it fails?  What does the wallet user see in this case, or it that a problem that the higher level client code needs to worry about and not core?

EBGToo: It is removed from the BRCryptoWallet, with events generated for higher code.

PR #429, opened 2020-01-17
CORE-799: Add TransferAttribute

EBGToo:
[Java work covered in CORE-807]  There are things to discuss about this Swift implementation (use of a value type for `TransferAttribute`) that might need to change after discussing Java implementation.  But, looking to get PR in place and willing to modify later.

### DIFF
--- a/crypto/BRCryptoTransfer.c
+++ b/crypto/BRCryptoTransfer.c
@@ -56,6 +56,8 @@ cryptoTransferCreateInternal (BRCryptoBlockChainType type,
     transfer->unitForFee = cryptoUnitTake(unitForFee);
     transfer->feeBasisEstimated = NULL;
 
+    array_new (transfer->attributes, 0);
+
     transfer->ref = CRYPTO_REF_ASSIGN (cryptoTransferRelease);
 
     {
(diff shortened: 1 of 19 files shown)


crypto/BRCryptoTransfer.c
```diff
@@ -416,6 +420,27 @@ cryptoTransferGetUnitForFee (BRCryptoTransfer transfer) {
     return cryptoUnitTake (transfer->unitForFee);
 }
 
+extern size_t
+cryptoTransferGetAttributeCount (BRCryptoTransfer transfer) {
+    return array_count(transfer->attributes);
```
michael-brd: There will be thread-sanitizer warnings here.

generic/BRGenericRipple.c
```diff
@@ -253,6 +272,49 @@ genericRippleWalletEstimateFeeBasis (BRGenericWalletRef wallet,
...
+    if (asRequired) { *count = rippleTransactionFieldRequiredCount; return rippleTransactionFieldRequiredNames; }
+    else {            *count = rippleTransactionFieldOptionalCount; return rippleTransactionFieldOptionalNames; }
+}
+
+static int
+genericRippleWalletValidateTransactionAttribute (BRGenericWalletRef wallet,
+                                                 BRGenericTransferAttribute attribute) {
+    if (0 == strcmp (attribute.key, "DestinationTag")) {
+        uint32_t tag;
+        return 1 == sscanf(attribute.value, "%u", &tag);
```
michael-brd: Might `attribute.value` be NULL?

crypto/BRCryptoTransfer.c
EBGToo: Fixed.  I don't see the BRCryptoWallet additions as needing a lock.

generic/BRGenericRipple.c
EBGToo: Fixed.

PR #430, opened 2020-01-17
CORE-809: Define known XRP addresses requiring 'DestinationTag'

EBGToo:


generic/BRGenericRipple.c
```diff
@@ -277,21 +277,70 @@ genericRippleWalletEstimateFeeBasis (BRGenericWalletRef wallet,
...
+    "rw2ciyaNshpHe7bCHo4bRWq6pqqynnWKQg",           // Coinbase(2)
+    "rEb8TK3gBgk5auZkwc6sHnwrGVJH8DuaLh",           // Binance(1)
+    "rJb5KsHsDHF1YS5B5DU6QCkH5NsPaKQTcy",           // Binance(2)
+    "rEy8TFcrAPvhpKrwyrscNYyqBGUkE9hKaJ",           // Binance(3)
+    "rXieaAC3nevTKgVu2SYoShjTCS2Tfczqx",            // Wirex(1)
+    "r9HwsqBnAUN4nF6nDqxd4sgP8DrDnDcZP3",           // BitBay
+    "rLbKbPyuvs4wc1h13BEPHgbFGsRXMeFGL6",           // BitBank(1)
+    "rw7m3CtVHwGSdhFjV4MyJozmZJv3DYQnsA",           // BitBank(2)
+    NULL
+};
```
carlcherry: Just one thought - should we be pulling this info from Blockset - just in case they add/change addresses.

EBGToo: I think that works too.  But we can be updating this list too, just like we update BTC, ETH, etc checkpoints w/ each Core release.

PR #434, opened 2020-01-24
CORE-804: Provide BlockSet metaData as 'TransferAttribute' data.

EBGToo:
[Currently built on CORE-805; I'll rebase once that PR is approved]

Needs a minimal change to update the Android UI for 'transfer detail' - add an 'Attributes' line.

### DIFF
--- a/Swift/BRCrypto/common/BRBlockChainDB.swift
+++ b/Swift/BRCrypto/common/BRBlockChainDB.swift
@@ -339,7 +339,8 @@ public class BlockChainDB {
             acknowledgements: UInt64,
             index: UInt64,
             transactionId: String?,
-            blockchainId: String)
+            blockchainId: String,
+            metaData: Dictionary<String,String>?)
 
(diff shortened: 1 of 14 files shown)


Swift/BRCrypto/common/BRBlockChainDB.swift
```diff
@@ -339,7 +339,8 @@ public class BlockChainDB {
             acknowledgements: UInt64,
             index: UInt64,
             transactionId: String?,
-            blockchainId: String)
+            blockchainId: String,
+            metaData: Dictionary<String,String>?)
```
michael-brd: Has this change been done in Java?

EBGToo: Yes, in `announceGetTransfersItemGen()`.  There is is destructures into its keys+values and passed onto Crypto C.

EBGToo:
I rebased onto staging/core-5.0.0

PR #435, opened 2020-01-27
CORE-811: Add null check for JNA indirect native calls

michael-brd:
@EBGToo, no longer sure of the branching strategy but this is a hotfix required by the Android guys. I've slatedit against staging/core-5.0.0 (and created a `core-5.0.0-beta.2.hotfix.1` tag for them as well). Let's chat.


EBGToo:
Just make this 'beta.3'

PR #447, opened 2020-02-18
CORE-839: `BRBase58DecodeEx()` can sometimes return '24'

EBGToo:


ripple/BRRippleAddress.c
```diff
@@ -148,7 +148,7 @@ BRRippleAddress rippleAddressStringToAddress(const char* input)
     uint8_t bytes[25];
     //int length = BRBase58DecodeEx(NULL, 0, input, rippleAlphabet);
     int length = rippleDecodeBase58(input, NULL);
-    if (length != 25) {
+    if (length > 25) {
```
michael-brd: We make some assumptions later on in this function about 25 bytes being filled in; are those still valid?

carlcherry: I ran a unit test with the ripple address that caused the crash - this change allowed us to go from ripple address (string) to ripple account (20 bytes) and then re-create the same string again - so it appears to be valid.  There is some bug in the BRBase58DecodeEx code which we will address soon.

EBGToo: Yes.  25 bytes are filled in - at least for the addresses that are causing problems.  Upon `uint8_t bytes[25]` I zeroed out the memory with `memset (bytes, 0, 25)` and then in the debugger examined `bytes` - confirmed that 25 are there.  The `memcmp()` succeeds (it would fail if `bytes[24]` was some stack-located random value). 

In addition Carl did a 'string->address' and then an 'address->string' (and visa-versa) and confirmed that the strings match.

At this point we have another JIRA (CORE-840) to investigate why `rippleDecodeBase58()` returns 24 when 25 bytes are actually provided.

michael-brd: Ah, so the old check was "right"; this change is a temporary workaround because the function is incorrectly returning 24? Cool.

EBGToo: Yes, exactly.

PR #448, opened 2020-02-20
CORE-816: Extend TransferConfirmation with success:Bool, error:String?

EBGToo:
Modify BRCryptoTransferState for 'included' to add 'BRCryptoBoolean success; char error[...]' and follow through for all of Crypto, Gen, Swift, Java.

The Java 'JNA Struct' seems to cause a crash.. on 'String error'.

### DIFF
--- a/Java/corenative/src/main/java/com/breadwallet/corenative/crypto/BRCryptoTransferState.java
+++ b/Java/corenative/src/main/java/com/breadwallet/corenative/crypto/BRCryptoTransferState.java
@@ -28,13 +28,14 @@ public class BRCryptoTransferState extends Structure {
         public errored_struct errored;
 
         public static class included_struct extends Structure {
+            private static int CRYPTO_TRANSFER_INCLUDED_ERROR_SIZE = 16;
 
             public long blockNumber;
             public long transactionIndex;
             public long timestamp;
             public BRCryptoFeeBasis feeBasis;
-            public int success;
-            public byte[] error = new byte[1];
+            public int success = BRCryptoBoolean.CRYPTO_TRUE;
+            public byte[] error = new byte[CRYPTO_TRANSFER_INCLUDED_ERROR_SIZE + 1];
 
             public included_struct() {
                 super();
--- a/Java/corenative/src/main/java/com/breadwallet/corenative/crypto/BRCryptoTransferState.java
+++ b/Java/corenative/src/main/java/com/breadwallet/corenative/crypto/BRCryptoTransferState.java
@@ -7,12 +7,15 @@
  */
 package com.breadwallet.corenative.crypto;
 
+import com.sun.jna.Native;
 import com.sun.jna.Pointer;
 import com.sun.jna.Structure;
 import com.sun.jna.Union;
 
 import java.util.Arrays;
 import java.util.List;
+
+import javax.annotation.Nullable;
 
 public class BRCryptoTransferState extends Structure {
 
@@ -30,8 +33,8 @@ public class BRCryptoTransferState extends Structure {
             public long transactionIndex;
             public long timestamp;
             public BRCryptoFeeBasis feeBasis;
-            public boolean success;
-            public String error;
+            public int success;
+            public byte[] error = new byte[1];
 
             public included_struct() {
                 super();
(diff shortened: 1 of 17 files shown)


EBGToo:
This change captures that a Transfer/Transaction can be on a blockchain, with a blockNumber and timestamp, but have failed.  Happens in XRP, apparently frequently.  And note, BlockSet DB provides this info as a status string "failed", "included", "submitted".  Default for non-GEN blockchains is 'success'.

EBGToo:
Also, need to consider persistent storage; format may need to change.

Java/corenative/src/main/java/com/breadwallet/corenative/crypto/BRCryptoTransferState.java
```diff
@@ -30,21 +30,25 @@
             public long transactionIndex;
             public long timestamp;
             public BRCryptoFeeBasis feeBasis;
+            public boolean success;
```
michael-brd: This should be `int`. This described the fields of the struct so must match the underlying C struct. Based on our recent discussion about boolean usage, this should be changed.

Java/crypto/src/main/java/com/breadwallet/crypto/TransferConfirmation.java
```diff
@@ -21,12 +21,16 @@
     private final UnsignedLong transactionIndex;
     private final Date timestamp;
     private final Optional<Amount> fee;
+    private final Boolean success;
```
michael-brd: Use `boolean` here, no?

Java/crypto/src/main/java/com/breadwallet/crypto/TransferConfirmation.java
```diff
@@ -21,12 +21,16 @@
...
+    private final Optional<String> error;
```
michael-brd: Pattern in the code base was to not use Optional for fields (but rather @Nullable annotate them). That said, did I use `Optional<Amount> fee` above?

Java/corenative/src/main/java/com/breadwallet/corenative/crypto/BRCryptoTransferState.java
```diff
@@ -30,21 +30,25 @@
...
+            public String error;
```
michael-brd: This should be public byte[] error = new byte[SIZE_GOES_HERE];

Then what I've done in the past (and could be done for `success` too once changed to an int), is have an accessor (`public boolean success()` and `public Optional<String> error()`) that takes the raw field and converts it to the more natural Java type.

EBGToo: Perfect.  Thanks.

Java/crypto/src/main/java/com/breadwallet/crypto/TransferConfirmation.java
EBGToo: This is a top-level, primary interface class.  `Optional` is appropriate here?

Java/crypto/src/main/java/com/breadwallet/crypto/TransferConfirmation.java
EBGToo: I changed it to `boolean` but the use of `UnsignedLong` lead me to use the class-based `Boolean` in the first place.

Java/corenative/src/main/java/com/breadwallet/corenative/crypto/BRCryptoTransferState.java
```diff
@@ -30,27 +33,39 @@
             public long transactionIndex;
             public long timestamp;
             public BRCryptoFeeBasis feeBasis;
+            public int success;
+            public byte[] error = new byte[1];
```
michael-brd: Isn't this (CRYPTO_TRANSFER_INCLUDED_ERROR_SIZE + 1)?

Java/corenative/src/main/java/com/breadwallet/corenative/crypto/BRCryptoTransferState.java
EBGToo: Done.

Java/corenative/src/main/java/com/breadwallet/corenative/crypto/BRCryptoTransferState.java
EBGToo: Yes, but I don't think it matters.  The constructor doesn't actually copying into that; it just assigns the fully-sized argument for `error`.

Java/crypto/src/main/java/com/breadwallet/crypto/TransferConfirmation.java
michael-brd: The guidance I had seen on Optional (and what the code base had been doing) was it was to be only used for return types (and local variables in functions).

If a field or parameter can be `null`, it is annotated with `@Nullable`

Java/crypto/src/main/java/com/breadwallet/crypto/TransferConfirmation.java
michael-brd: Yeah, the UnsignedLong is because Java doesn't have that type natively.

Java/corenative/src/main/java/com/breadwallet/corenative/crypto/BRCryptoTransferState.java
michael-brd: The JNA docs indicate that the size of the structure is calculated based off of this, when you have an array within a structure.

michael-brd: In https://java-native-access.github.io/jna/4.2.1/overview-summary.html#structures, search for:
Structures with nested arrays require an explicit constructor to ensure the structure size is properly calculated.

EBGToo: It seems the code does not go through the 'constructor' `public included_struct(...)` - so, in fact, I am only getting the first character of the error.

EBGToo:
I updated the persistent format for GEN transfers from 'v0' to 'v1'.  I confirmed that 'v0' storage got updated to 'v1' storage.  I confirmed that the BlocksetDB query result updated the transfer to 'included w/ error' and they got displayed in the correct order.

Testing 'airplane mode'...

EBGToo:
Confirmed w/ airplane mode.  Using current IOS version, upgrade (to this code), start the App in airplane mode - failed XRP transaction are shown as 'failed w/o being included'.  Then exit airplane mode, BlockSet is queried, XRP transactions changed to 'included but e/ error'.

PR #452, opened 2020-03-18
CORE-853: check tx signature/witness for wallet inclusion

EBGToo:
Also modifies BRWalletManager.c to address possible P2P issues seen in IOS related to 'isDeleted'

### DIFF
--- a/bitcoin/BRWalletManager.c
+++ b/bitcoin/BRWalletManager.c
@@ -563,11 +563,12 @@ BRWalletManagerFindTransactionWithLastConfirmedSend(BRWalletManager manager,
  */
 static BRTransactionWithState
 BRWalletManagerFindTransactionByHash (BRWalletManager manager,
-                                      UInt256 hash) {
+                                      UInt256 hash,
+                                      int ignoreIsDeleted) {
     BRTransactionWithState txnWithState = NULL;
 
     for (size_t index = 0; index < array_count(manager->transactions); index++) {
-        if (!manager->transactions[index]->isDeleted &&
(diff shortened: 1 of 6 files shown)


EBGToo:
@voisine Can you review `BRWalletIsResolved` (and associated `_BRWalletContainsTxInput`)?  A transaction is resolved if all inputs on wallet have known dependent transactions.

@michael-brd Can you review again? I added an 'isResolved' flag to BRTransactionWithState.  Until resolved events are not generated.  In 'txAdded' all known BRWalletManager transactions are reviewed; any that are newly resolved have their {ADDED, UPDATED} events generated.

EBGToo:
I've tested to show that API and P2P modes both sync successfully with the 'transaction that arrives early' not being announced until its dependent transaction arrives.  Aaron's earlier change that recognized the 'transaction that arrives early' adds the transaction to wallet; the 'isResolved' change avoids that added transaction from having a zero amount.

bitcoin/BRWalletManager.c
```diff
@@ -1364,6 +1364,7 @@ BRWalletManagerCreateTransaction (BRWalletManager manager,
     pthread_mutex_unlock (&manager->lock);
 
     if (NULL != txnWithState) {
+        BRTransactionWithStateSetResolved (txnWithState);  // Always resolved if created
```
michael-brd: Modifications should be in the lock, I believe. To be honest, the gets should be there too (a mistake on my part).

EBGToo: Addressed w/ the 'gets' too.

EBGToo:
Tested sending a transaction.  As expected 'isResolved' and events are generated.

PR #475, opened 2021-03-02
Check outputs more than dust threshold

Bushstar:
A minAmount is calculated, 5,460 Sats, this is being added to the total amount required for the inputs if the amount selected does not meet the amount required plus fees exactly. From the comment on BRWalletMinOutputAmountWithFeePerKb it would seem that it is designed to check output amounts and not increase the amount that the inputs need to cover.

The PR here checks inputs against the "dust" amount, it will return 546 Sats, outputs are then checked against this. If we expect external systems to check this and do not want to sanity check it then we can remove minAmount entirely.

minAmount is then removed as a requirement for the inputs balance when the exact amount of fees + output balance is not met.

Perhaps I misunderstand the logic here. Please let me know and I'll update or close as required. I'm working with dust inputs and with the minAmount being added when the exact amount is not found I then need an extra ten inputs to cover that amount and then more to cover fees, this is somewhat problematic.

### DIFF
--- a/bitcoin/BRWallet.c
+++ b/bitcoin/BRWallet.c
@@ -635,13 +635,19 @@ BRTransaction *BRWalletCreateTxForOutputsWithFeePerKb(BRWallet *wallet, uint64_t
     assert(wallet != NULL);
     assert(outputs != NULL && outCount > 0);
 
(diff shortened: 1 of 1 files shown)


Bushstar:
I get it now, it is for the change output to make sure it is at least 5,460 Sats!

